PR #5, opened 2019-08-23
Add nixpkgs-fmt, Nix formatter written in Rust

domenkozar:
cc @zimbatm


nix/default.nix
```diff
@@ -21,6 +21,7 @@ with {
           inherit (import sources.niv {}) niv;
           inherit (import sources.ormolu {}) ormolu;
           inherit (import sources.canonix {}) canonix;
+          nixpkgs-fmt = import sources.nixpkgs-fmt {};
```
zimbatm: I need to test if this works. It would be nice not to instantiate nixpkgs twice:
```suggestion
          nixpkgs-fmt = import sources.nixpkgs-fmt { inherit pkgs; };
```
I also need to fix the Cargo sha256.

zimbatm: The inherit should work if you use nixos-unstable. The version of rust in nixos-19.03 is too old though.

I also fixed the release process in master to keep the nix sha256 in sync.

PR #7, opened 2019-09-03
Fix shellHook PATH

roberth:
- [x] test


domenkozar:
I forgot I have it globally installed, so my testing was futile :)

PR #10, opened 2019-09-05
Add Terraform

roberth:
cc @zimbatm


domenkozar:
terraform-docs-updater should be maintained separately (nix-community?), this repo is just about pulling things together.

roberth:
I agree, although about a third of the bash code will have to stay here because that's glue to hook directory-based formatters up to a file-based incremental system, fwiw.

I've looked for a better solution, but couldn't find anything except foregoing incrementality altogether with the unintended side effect of having to rewrite the recursion part. I'd love to be proven wrong on this one :man_shrugging:

domenkozar:
I'm fine having the file to folder conversion here as that's core to how pre-commit works.

But terraform-doc bash should go to a separate repository so it can be maintained there, with a README and it's own write users list. 

I really don't want to encourage more people to just add bash scripts here, and as soon as the first one lands, it's going to be hard to defend that.

roberth:
The docs updater may be replaced by a tool feature like https://github.com/zimbatm/mdsh/issues/24

zimbatm:
Right, mdsh already has support for pre-commit hooks as well. Currently it only updates the top-level README but it could be extended.

roberth:
@domenkozar this only adds `terraform fmt` now

PR #16, opened 2019-09-17
Add hindent hook

### LINKED ISSUE #15: Hindent as well.
Ormolu is currently supported but it doesn't look like hindent is.
I don't find ormolu mature enough to use, and I'm using hindent already, so I'd like to set up a hook for hindent.

NorfairKing:
This should fix #15 but I'm getting a strange error while using it:

```
hindent..................................................................Failed
hookid: hindent

Executable `/nix/store/ffg7l3wannl3vi5ijyv533p43bmkb50r-ormolu-0.0.1.0/bin/hindent` not found

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ eval "$shellHook"
 ## Haskell
 
 - [ormolu](https://github.com/tweag/ormolu)
+- [hindent](https://github.com/chrisdone/hindent)
 - [hlint](https://github.com/ndmitchell/hlint)
 - [cabal-fmt](https://github.com/phadej/cabal-fmt)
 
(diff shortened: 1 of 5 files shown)


domenkozar:
bors try

domenkozar:
Once you bumped nix-pre-commit-hooks, did you re-enter nix-shell?

NorfairKing:
@domenkozar Yes.
The error happens when I run `nix-build -A pre-commit-check`, not when committing.
And the error is also strange, look at the hash:

```
Executable `/nix/store/ffg7l3wannl3vi5ijyv533p43bmkb50r-ormolu-0.0.1.0/bin/hindent` not found
```

NorfairKing:
Oooh, I forgot to update the sha256

domenkozar:
bors try

NorfairKing:
Woop woop :fireworks: 
Now I just have to somehow get my local system to agree with the pre-commit hooks on the hindent version to use, because they seem to disagree.

PR #18, opened 2019-09-22
Fix nix 2.3

roberth:
Inherit the gitignore fix for canonix (prevented eval of this project)


roberth:
GitHub review enforcement was a problem for bors. I've turned it off.

PR #19, opened 2019-09-25
Use the module system

### LINKED ISSUE #2: Add docs, examples for overriding provided hooks
Something like:

```nix
nix-pre-commit-hooks.run {
  src = gitignoreSource ./.;
  hooks = ["shellcheck"];
}
```

roberth:
Closes #2
Closes #17 

TODO
 - [x] ~expose and test "core" module set from project.nix~
 - [x] ~niv update -b master project.nix when that's done (https://github.com/hercules-ci/project.nix/pull/10)~
 - [x] changelog entry: (elaborate on) translate `.pre-commit-config.yaml` to hooks argument
 - [x] update README
 - [x] add how to add a hook to README

New
 - Selectively enable tools (#2)
 - Make choice of tool versions overridable
 - User-definable hooks
 - Input validation at the Nix level
 - Use gitignore by default

Fixes
 - Stop installing files all over the place when doing nix-shell ../some-project to borrow a tool
 - Lorri support

Cleanup
 - Get rid of some installation code


README.md
```diff
@@ -111,6 +104,5 @@ Everyone is encouraged to add new hooks.
 
 There's no guarantee the hook will be accepted, but the general guidelines are:
 
-- Nix closure of the tool should be small e.g. `< 50MB`
```
domenkozar: I think this should stay, even if it's currently not true for existing tooling.

roberth:
> There are still references to project.nix, which are not used.

It's used by the `shellHook` for the auto-installation part. With "delegate" in the PR title I meant pulling it in from there and maintaining it there. Removing `shellHook` entirely and directing people to project.nix for this functionality seems a bit too soon.
So do we want this project to be usable on its own? If so, does it matter whether we pin or duplicate?

domenkozar:
> Removing shellHook entirely and directing people to project.nix for this functionality seems a bit too soon.

Why can't we just expose shellHook as it was before this PR and project.nix can pick it up?

Maybe there should be `nix/project.nix` with:

```
{ ... }: {
  config.activation.hooks = (import ../default.nix { }).shellHook;
}
```

Which is later going to be translated into flakes.

Once we need to merge project.nix options we can do that.

roberth:
> Why can't we just expose shellHook as it was before this PR

This PR does not change that part of the nix-pre-commit API, so it is exposed as before, but with an implementation that is not directly inside the repo. The project.nix implementation fixes some problems and has improvements over the current shellHook implementation. Of course these can be backported but that seems a bit pointless to me.

> and project.nix can pick it up?

This is can be done, but it must be an acceptable hook for the [activation module](https://github.com/hercules-ci/project.nix/pull/10/files#diff-cce804161f751d923222a012ea088d75), rather than a standalone autoinstall script. It's not a big difference, just have to expose two things instead of one.

roberth:
bors is broken?
Merging as admin...

Anyway :tada:

PR #21, opened 2019-10-05
Make pre-commit.tools lazily checked

roberth:
This results in a two-fold performance improvement in

    time nix-instantiate shell.nix

`lazyAttrsOf.nix` will be a polyfill when the thing is upstreamed.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -12,6 +12,7 @@ let
     mkOption
     types
     ;
+  inherit (import ../nix/lazyAttrsOf.nix { inherit lib; }) lazyAttrsOf;
 
   inherit (pkgs) runCommand writeText git;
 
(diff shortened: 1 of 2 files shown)


domenkozar:
5s -> 2.5s looks good!

roberth:
Looks like we need to configure github not to require a review for bors to work well, or always review through github first.

domenkozar:
The required check should be on `ci/hercules`, not `bors`.

roberth:
Not sure. There's some value in always going through bors, so requiring it may make sense. Bors is reporting a problem with reviews, not statuses. I guess whether bors can be a required check hinges on whether it sets a status before merging. I expect that it does.

domenkozar:
> Bors is reporting a problem with reviews

It piggybacks on ci/hercules status after merging the code, so it's essentially the same. What else does it do?

roberth:
> It piggybacks on ci/hercules status after merging the code, so it's essentially the same.

But not quite the same. It's the result after merge that we care about. Its the difference between a branch status and a pr status. Insignificant most of the time, until you break master because of the difference.

roberth:
It just keeps saying failed, without any reasons. :( Will do a noop commit amend and force push now.

roberth:
Merged while required+up-to-date bors status was enabled.

PR #22, opened 2019-10-05
Global exclude, restore/update sources.nix

roberth:


nix/packages.nix
```diff
@@ -15,5 +15,8 @@ in
       shellcheck.enable = true;
       canonix.enable = true;
     };
+    excludes = [
+      "nix/sources.nix$"
```
domenkozar:
```suggestion
      # autogenerated by niv
      "nix/sources.nix$"
```

PR #25, opened 2019-11-06
Fix caching

roberth:
I hope this is acceptable because it already took too many iterations imo.
:rabbit: 
:hole:


domenkozar:
Seems good to me - I'd only open an upstream issue and link to our solution to get the discussion going.

PR #30, opened 2019-11-23
Add ansible-lint

PierreR:
I have tested this PR successfully.

How would this ansible-linter interacts with a yaml linter ? For instance I have a project where ansible files are in a folder 'ansible' while other yaml file should be linted by the yaml linter.

I know the yaml linter is not yet there but hopefully it will very soon.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -104,5 +104,11 @@ in {
           entry = "${tools.terraform-fmt}/bin/terraform-fmt";
           files = "\\.tf$";
         };
+      yamllint =
+        {
+          name = "yamllint";
+          description = "Yaml linter";
+          entry = "${tools.yamllint}/bin/yamllint";
+        };
     };
 }
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -24,7 +24,6 @@ in {
           description =
             "Ansible linter";
           entry = "${tools.ansible-lint}/bin/ansible-lint";
-          files = "\\.yml$";
         };
       hlint =
         {
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -18,6 +18,14 @@ in {
 
   config.pre-commit.hooks =
     {
+      ansible-lint =
+        {
+          name = "ansible-lint";
+          description =
+            "Ansible linter";
+          entry = "${tools.ansible-lint}/bin/ansible-lint";
+          files = "\\.yml$";
+        };
       hlint =
         {
           name = "hlint";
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -32,7 +40,7 @@ in {
           description = "Haskell code prettifier.";
           entry =
             "${tools.ormolu}/bin/ormolu --mode inplace ${
-              lib.escapeShellArgs (lib.concatMap (ext: ["--ghc-opt" "-X${ext}"]) settings.ormolu.defaultExtensions)
+            lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) settings.ormolu.defaultExtensions)
```
PierreR: This bit was done by nixpkgs-fmt, Let me know if it is better to remove it.

modules/hooks.nix
```diff
@@ -18,6 +18,14 @@ in {
 
   config.pre-commit.hooks =
     {
+      ansible-lint =
+        {
+          name = "ansible-lint";
+          description =
+            "Ansible linter";
+          entry = "${tools.ansible-lint}/bin/ansible-lint";
+          files = "\\.yml$";
```
PierreR: Technically it should be "\\.(yaml|yml)$" but I am not sure what is possible here ?

roberth: I think this should be left to the user, to point it at the right files that are actually ansible files.
The [ansible's .pre-commit-hooks.yaml](https://github.com/ansible/ansible-lint/blob/master/.pre-commit-hooks.yaml) also leaves it out.

modules/hooks.nix
```diff
@@ -96,5 +104,11 @@ in {
           entry = "${tools.terraform-fmt}/bin/terraform-fmt";
           files = "\\.tf$";
         };
+      yamllint =
+        {
+          name = "yamllint";
+          description = "Yaml linter";
+          entry = "${tools.yamllint}/bin/yamllint";
```
roberth: This one seems generic enough to use a pattern that matches all yaml files. Upstream also [does this](https://github.com/adrienverge/yamllint/blob/master/.pre-commit-hooks.yaml).

PierreR: That's the reason why I didn't specify the 'files' properties.
Is there any left to do to improve the PR ?
Thx for the review.

roberth: Ah so I was expecting something like `types = ["file" "yaml"]` here.
After enabling, users can still opt out for some set of files via the `excludes` options.

PierreR: Apply in c28ce2cd8a8ce1b4e2761f61da10dd79cbd5b8aa.
I have to admit I don't quite get the syntax ... Are the two elements of the list treated as key/value ?

roberth: It's what pre-commit does. Apparently it uses [this mapping](https://github.com/chriskuehl/identify/blob/master/identify/extensions.py). How [types](https://pre-commit.com/#filtering-files-with-types) are combined is left implicit in the docs, but it needs to be set intersection / logical and to make sense.

roberth:
bors try

roberth:
Thanks a lot PierreR!

bors r+

PR #33, opened 2020-01-07
Add nix-linter

anpryl:
Tested on my projects

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -15,6 +15,16 @@ in
               default = [];
             };
         };
+      nix-linter =
+        {
+          checks =
+            mkOption {
+              type = types.listOf types.str;
+              description =
+                "Available checks (See `nix-linter --help-for [CHECK]` for more details)";
+              default = [];
+            };
+        };
     };
 
   config.pre-commit.hooks =
@@ -70,6 +80,16 @@ in
           name = "nixpkgs-fmt";
           description = "Nix code prettifier.";
           entry = "${tools.nixpkgs-fmt}/bin/nixpkgs-fmt";
+          files = "\\.nix$";
+        };
+      nix-linter =
+        {
+          name = "nix-linter";
+          description = "Linter for the Nix expression language.";
+          entry =
+            "${tools.nix-linter}/bin/nix-linter ${
+            lib.escapeShellArgs (lib.concatMap (check: [ "-W" "${check}" ]) settings.nix-linter.checks)
+            }";
           files = "\\.nix$";
         };
       elm-format =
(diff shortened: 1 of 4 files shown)


domenkozar:
Any way to turn all by default?

anpryl:
By default it runs all checks except:
```
    AlphabeticalArgs (disabled by default)
    AlphabeticalBindings (disabled by default)
    BetaReduction (disabled by default)
    EmptyVariadicParamSet (disabled by default)
    UnneededAntiquote (disabled by default)
```

I can add an additional boolean option to enable all checks.
Do you like this idea?

modules/hooks.nix
```diff
@@ -14,6 +14,16 @@ in {
               default = [];
             };
         };
+      nix-linter =
+        {
+          checks =
+            mkOption {
+              type = types.listOf types.str;
+              description =
+                "Available checks (See `nix-linter --help-for [CHECK]` for more details)";
```
domenkozar:
```suggestion
                "Available checks (See `nix-linter --help-for [CHECK]` for more details). See https://github.com/Synthetica9/nix-linter#usage for what checks are enabled by default.";
```

domenkozar: @anpryl how about this? :)

modules/hooks.nix
```diff
@@ -14,6 +14,16 @@ in {
               default = [];
             };
         };
+      nix-linter =
+        {
+          checks =
+            mkOption {
+              type = types.listOf types.str;
```
roberth: I generally prefer `types.attrsOf types.bool` because it lets you both enable and disable, and you can then override individual rules with `lib.mkForce`.

anpryl: Do you propose to have a record with all available extensions as fields an bool flags to enable/disable it?

domenkozar: @anpryl I think he meant it to be free form.

domenkozar:
Could you rebase on top of master? :)

anpryl:
Sorry for delay, will work on this this week.

anpryl:
Rebased on master, but I want to work on https://github.com/cachix/pre-commit-hooks.nix/pull/33#discussion_r364764001

domenkozar:
Great :) Make sure to add it to the README

nix/sources.json
```diff
@@ -59,6 +59,18 @@
         "url": "https://github.com/nmattia/niv/archive/efce82e4ba77a84835f9febefe3b2fe52a8c7ede.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
+    "nix-linter": {
```
domenkozar, line 62: Any reason not to use it from nixpkgs instead?

anpryl: I didn't find it in nixpkgs on initial implementation.
I can update PR.

domenkozar:
As this is already useful I'll merge it, if we change the types that seems like a reasonable backwards incompatible change that types are there to enforce :-)

PR #50, opened 2020-07-07
Avoid unexpected rebuilds

blaggacao:
See: https://github.com/target/lorri#lorri-reevaluates-more-than-expected
And: https://github.com/target/lorri/issues/6#issuecomment-478288372


blaggacao:
It doesn't work.

domenkozar:
It still lags lorri?

blaggacao:
I _think_ this actually hasn't been a problem of `default.nix` and pre-commit. 

I observed it where I had a ref to `${./.}` in my `looriHook` in `shell.nix`.

Once I fixed that, it didn't rebuild any more, even though pre-commit-nix still hold a ref to `./.` in `default.nix`.

This is tricky, maybe `lorri` should find a more prominent way of dealing with this issue.

PR #52, opened 2020-07-20
feat: support specifying stages

### LINKED ISSUE #49: Support specifying stages
For speedy development, I'd configure certain hooks for different stages.

A common example:

- run pre-commit hooks on the `prepush` stage (to make sure only "clean" stuff gets remoted)
- don't run anything on _commit_ to avoid disrupting _developer's **flow**_
- alias `prepush` hooks from a justfile, so developer can run - e.g. - `just fmt` conscientiously and avoid surprises at `git push` (promote the stimulus of being in control)

Therefore, I'd need to configure:
- either `hook = { stages = ["manual", "push"] }`
- or something global `... run { ...; stages = [ "manual", "push" ] }`

blaggacao:
closes #49

Since I'm a newcomer, I have no way to judge (or even validate) if that's the right way _at all_: it's a shot into the _darknix_.
If it's a hit, I appreciate that I could seamlessly contribute. If it's a miss, please guild me around to get this merge-ready.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -350,7 +350,7 @@ in
                         pre-commit install -t $stage
                         ;;
                       *)
-                        echo 1>&2 "ERROR: nix-pre-commit-hooks: either $stage is not a valid stage or pre-commit-hook.nix doesn't yet support it."
+                        echo 1>&2 "ERROR: nix-pre-commit-hooks: either $stage is not a valid stage or pre-commit-hooks.nix doesn't yet support it."
                         exit 1
                         ;;
                     esac
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -331,24 +331,14 @@ in
                 ln -s ${configFile} .pre-commit-config.yaml
                 # Remove any previously installed hooks (since pre-commit itself has no convergent design)
                 hooks="pre-commit pre-merge-commit pre-push prepare-commit-msg commit-msg post-checkout post-commit"
-                uninstall_help=$(pre-commit uninstall --help)
-                install_help=$(pre-commit install --help)
                 for hook in $hooks; do
-                  # different version of pre-commit support different hooks
-                  if [[ "$uninstall_help" =~ "$hook" ]]; then
-                    pre-commit uninstall -t $hook
-                  fi
+                  pre-commit uninstall -t $hook
                 done
                 # Add hooks for configured stages (only) ...
                 if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
                   for stage in ${concatStringsSep " " cfg.default_stages}; do
                     if [[ "$stage" == "manual" ]]; then
                       continue
-                    fi
-                    # different version of pre-commit support different hooks
-                    if [[ ! "$install_help" =~ "$stage" ]];then
-                        echo 1>&2 "WARNING: nix-pre-commit-hooks: '$stage' is not supported by this pre-commit version."
-                        continue
                     fi
                     case $stage in
                       commit | merge-commit | push)
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -258,6 +258,18 @@ in
           default = [];
         };
 
+      default_stages =
+        mkOption {
+          type = types.listOf types.str;
+          description =
+            ''
+              A configuration wide option for the stages property.
+              Installs hooks to the defined stages.
+              Default is empty which falls back to 'commit'.
+            '';
+          default = [];
+        };
+
       rawConfig =
         mkOption {
           type = types.attrs;
@@ -286,6 +298,8 @@ in
             ];
         } // lib.optionalAttrs (cfg.excludes != []) {
           exclude = mergeExcludes cfg.excludes;
+        } // lib.optionalAttrs (cfg.default_stages != []) {
+          default_stages = cfg.default_stages;
         };
 
       pre-commit.installationScript =
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -349,6 +349,8 @@ in
                         pre-commit install -t $stage
                         ;;
                       *)
+                        echo 1>&2 "ERROR: nix-pre-commit-hooks: either $stage is not a valid stage or pre-commit-hook.nix doesn't yet support it."
+                        exit 1
                         ;;
                     esac
                   done
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -17,6 +17,8 @@ let
   inherit (pkgs) runCommand writeText git;
 
   cfg = config.pre-commit;
+  # TODO: add per-hook stages to install here
+  install_stages = cfg.default_stages;
 
   hookType =
     types.submodule (
@@ -335,8 +337,8 @@ in
                   pre-commit uninstall -t $hook
                 done
                 # Add hooks for configured stages (only) ...
-                if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
-                  for stage in ${concatStringsSep " " cfg.default_stages}; do
+                if [ ! -z "${concatStringsSep " " install_stages}" ]; then
+                  for stage in ${concatStringsSep " " install_stages}; do
                     if [[ "$stage" == "manual" ]]; then
                       continue
                     fi
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -329,7 +329,31 @@ in
                 echo 1>&2 "    3. add .pre-commit-config.yaml to .gitignore"
               else
                 ln -s ${configFile} .pre-commit-config.yaml
-                pre-commit install
+                # Remove any previously installed hooks (since pre-commit itself has no convergent design)
+                hooks="pre-commit pre-merge-commit pre-push prepare-commit-msg commit-msg post-checkout post-commit"
+                for hook in $hooks; do
+                  pre-commit uninstall -t $hook
+                done
+                # Add hooks for configured stages (only) ...
+                if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
+                  for stage in ${concatStringsSep " " cfg.default_stages}; do
+                    case $stage in
+                      commit | merge-commit | push)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
+                      prepare-commit-msg | commit-msg | post-checkout | post-commit)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
+                      *)
+                        ;;
+                    esac
+                  done
+                # ... or default 'pre-commit' hook
+                else
+                  pre-commit install
+                fi
               fi
             fi
           fi
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -341,6 +341,7 @@ in
                       continue
                     fi
                     case $stage in
+                      # if you amend these switches please also review $hooks above
                       commit | merge-commit | push)
                         stage="pre-"$stage
                         pre-commit install -t $stage
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -331,19 +331,31 @@ in
                 ln -s ${configFile} .pre-commit-config.yaml
                 # Remove any previously installed hooks (since pre-commit itself has no convergent design)
                 hooks="pre-commit pre-merge-commit pre-push prepare-commit-msg commit-msg post-checkout post-commit"
+                uninstall_help=$(pre-commit uninstall --help)
+                install_help=$(pre-commit install --help)
                 for hook in $hooks; do
-                  pre-commit uninstall -t $hook
+                  # different version of pre-commit support different hooks
+                  if [[ "$uninstall_help" =~ "$hook" ]]; then
+                    pre-commit uninstall -t $hook
+                  fi
                 done
                 # Add hooks for configured stages (only) ...
                 if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
                   for stage in ${concatStringsSep " " cfg.default_stages}; do
+                    if [[ "$stage" == "manual" ]]; then
+                      continue
+                    fi
+                    # different version of pre-commit support different hooks
+                    if [[ ! "$install_help" =~ "$stage" ]];then
+                        echo 1>&2 "WARNING: nix-pre-commit-hooks: '$stage' is not supported by this pre-commit version."
+                        continue
+                    fi
                     case $stage in
                       commit | merge-commit | push)
                         stage="pre-"$stage
                         pre-commit install -t $stage
                         ;;
                       prepare-commit-msg | commit-msg | post-checkout | post-commit)
-                        stage="pre-"$stage
                         pre-commit install -t $stage
                         ;;
                       *)
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -46,17 +46,5 @@
         "type": "tarball",
         "url": "https://github.com/NixOS/nixpkgs/archive/99a3d7a86fce9e9c9f23b3e304d7d2b1270a12b8.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
-    },
-    "pre-commit": {
-        "branch": "v2.6.0",
-        "description": "A framework for managing and maintaining multi-language pre-commit hooks.",
-        "homepage": "https://pre-commit.com",
-        "owner": "pre-commit",
-        "repo": "pre-commit",
-        "rev": "6fe1702ee106f4d7f4a8ad73550db2145208ef24",
-        "sha256": "144hcnz8vz07nkx7hk8a3ac822186ardwxa8jnl6s8qvm5ip92f2",
-        "type": "tarball",
-        "url": "https://github.com/pre-commit/pre-commit/archive/6fe1702ee106f4d7f4a8ad73550db2145208ef24.tar.gz",
-        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     }
 }
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -36,15 +36,27 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "nixpkgs": {
-        "branch": "nixpkgs-unstable",
-        "description": "Nix Packages collection",
-        "homepage": null,
+        "branch": "nixos-20.03",
+        "description": "A read-only mirror of NixOS/nixpkgs tracking the released channels. Send issues and PRs to",
+        "homepage": "https://github.com/NixOS/nixpkgs",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "2d6cbbe4627f6fe4a179c681537b0a3e4f59b732",
-        "sha256": "12kh498rfr0yscvzidbgvip3x2zah098nsi6mjj3vfrlz2aximi5",
+        "rev": "99a3d7a86fce9e9c9f23b3e304d7d2b1270a12b8",
+        "sha256": "0i40cl3n6600z2lkwrpiy28dcnv2r63fcgfswj91aaf1xfn2chql",
         "type": "tarball",
-        "url": "https://github.com/NixOS/nixpkgs/archive/2d6cbbe4627f6fe4a179c681537b0a3e4f59b732.tar.gz",
+        "url": "https://github.com/NixOS/nixpkgs/archive/99a3d7a86fce9e9c9f23b3e304d7d2b1270a12b8.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "pre-commit": {
+        "branch": "v2.6.0",
+        "description": "A framework for managing and maintaining multi-language pre-commit hooks.",
+        "homepage": "https://pre-commit.com",
+        "owner": "pre-commit",
+        "repo": "pre-commit",
+        "rev": "6fe1702ee106f4d7f4a8ad73550db2145208ef24",
+        "sha256": "144hcnz8vz07nkx7hk8a3ac822186ardwxa8jnl6s8qvm5ip92f2",
+        "type": "tarball",
+        "url": "https://github.com/pre-commit/pre-commit/archive/6fe1702ee106f4d7f4a8ad73550db2145208ef24.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     }
 }
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -46,5 +46,18 @@
         "type": "tarball",
         "url": "https://github.com/NixOS/nixpkgs/archive/99a3d7a86fce9e9c9f23b3e304d7d2b1270a12b8.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "pre-commit": {
+        "branch": "master",
+        "description": "A framework for managing and maintaining multi-language pre-commit hooks.",
+        "homepage": "https://pre-commit.com",
+        "owner": "pre-commit",
+        "repo": "pre-commit",
+        "rev": "c0f750d9b36998be3090476cc4b33cde9b9e65c3",
+        "sha256": "1l018cqs8jsjyqhmfas9qckakjgzck0wl9w98170nk08arllsa2q",
+        "type": "tarball",
+        "url": "https://github.com/pre-commit/pre-commit/archive/c0f750d9b36998be3090476cc4b33cde9b9e65c3.tar.gz",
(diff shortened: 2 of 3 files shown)


roberth:
> *darknix*

nice :D

I think it's missing the option declaration. I haven't read the pre-commit docs so maybe it's not accurate, but perhaps `excludes` can serve as an example https://github.com/cachix/pre-commit-hooks.nix/pull/52/files#diff-0920dde2a9dc0ea9c564d0c0131c8feeR251

You can test your change by using it in a project. We don't have a test suite yet.

roberth:
You can replace the `builtins.fetchTarbal "..."` part in the README by an unquoted local path like `../pre-commit.nix` for testing

blaggacao:
Thanks! I broke things. :cry: 

<details>

```console
nix-shell --show-trace
warning: unknown setting 'experimental-features'
unpacking 'https://github.com/blaggacao/pre-commit-hooks.nix/tarball/master'...
error: while evaluating the attribute 'lorriHook' of the derivation 'nix-shell' at /nix/store/isyka0a95cxhj1381bg3bbqdv78qrln9-nixpkgs-20.09pre235189.d7e20ee25ed/nixpkgs/pkgs/build-support/mkshell/default.nix:28:3:
while evaluating the attribute 'pre-commit-check.shellHook' at /home/blaggacao/ghq/git.p1.rocks/gitops/aedir-ldap.k8s/default.nix:6:3:
while evaluating anonymous function at /nix/store/b4jq885hsxy31h23f3b6c5dnizchgn5q-source/nix/run.nix:3:1, called from /home/blaggacao/ghq/git.p1.rocks/gitops/aedir-ldap.k8s/default.nix:6:22:
while evaluating the attribute 'config.pre-commit.run' at undefined position:
while evaluating anonymous function at /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:84:45, called from undefined position:
while evaluating 'yieldConfig' at /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:83:29, called from /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:86:16:
while evaluating anonymous function at /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:89:24, called from /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:89:11:
while evaluating anonymous function at /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:90:26, called from /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/lib/modules.nix:90:13:
The option `pre-commit.default_stages' defined in `<unknown-file>' does not exist.
```

</details>

roberth:
Oh my link above is broken. You need to declare an option for `default_stages` with `mkOption`. It's using the NixOS module system.

blaggacao:
Now it builds, but the option still is a no show in `.pre-commit-config.yaml` ?

modules/pre-commit.nix
```diff
@@ -258,6 +258,18 @@ in
...
         };
 
+      default_stages =
+        mkOption {
+          type = types.listOf types.str;
+          description =
+            ''
+              A configuration wide option for the stages property.
+              Installs hooks to the defined stages.
+              Default is empty which falls back to 'commit'.
```
blaggacao, line 268: Is this wording ok?

blaggacao, line 268: Is this wording ok?

blaggacao:
I do fake `repo = "locall";` (double `l`) and it doesn't change, neither. ?!?

blaggacao:
Never mind, it was `lorri`'s fault (or mine, however you look at it - I prefer to blame `lorri`). This PR is good to go.

blaggacao:
<details>
<summary>Example result</summary>

```
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: .pre-commit-config.yaml
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ # DO NOT MODIFY
   2   │ # This file was generated by nix-pre-commit-hooks
   3   │ {
   4   │   "default_stages": [
   5   │     "manual",
   6   │     "pull"
   7   │   ],
   8   │   "exclude": "(^k8s/kustomize/|^images/ansible-ae-dir-server/|^images/config-initializer/ae-dir-configuration.sample.yml$|^images/config-initializer/ae-dir-customization.yml$|^images/config-initializer/ae-dir-container.yml$)",
   9   │   "repos": [
  10   │     {
  11   │       "hooks": [
  12   │         {
  13   │           "entry": "/nix/store/2pwj46x0qq9asnxkag1qcbvkz2zf2a9w-kubeval-0.15.0/bin/kubeval",
  14   │           "exclude": "^$",
  15   │           "files": "^k8s/.*?\\.(yaml|yml)$",
  16   │           "id": "kubeval",
  17   │           "language": "system",
  18   │           "name": "kubeval",
  19   │           "pass_filenames": true,
  20   │           "types": []
  21   │         },
  22   │         {
  23   │           "entry": "/nix/store/fisrj8c6cizy9df8hbg9cz1y64yvb1bv-node_prettier-2.0.5/bin/prettier --write --list-different",
  24   │           "exclude": "^$",
  25   │           "files": "\\.(css|less|scss|graphql|gql|html|js|jsx|json|md|markdown|mdown|mkdn|mdx|ts|tsx|vue|yaml|yml)$",
  26   │           "id": "prettier",
  27   │           "language": "system",
  28   │           "name": "prettier",
  29   │           "pass_filenames": true,
  30   │           "types": []
  31   │         },
  32   │         {
  33   │           "entry": "/nix/store/000llwy4g16jhmbi3yzngyzp1k4qykp0-ShellCheck-0.7.0/bin/shellcheck",
  34   │           "exclude": "^$",
  35   │           "files": "",
  36   │           "id": "shellcheck",
  37   │           "language": "system",
  38   │           "name": "shellcheck",
  39   │           "pass_filenames": true,
  40   │           "types": [
  41   │             "bash"
  42   │           ]
  43   │         },
  44   │         {
  45   │           "entry": "/nix/store/zihp6pxl6s8cb45zfbn2hcqkhis2zsnf-shfmt-3.1.2/bin/shfmt -w -l",
  46   │           "exclude": "^$",
  47   │           "files": "",
  48   │           "id": "shfmt",
  49   │           "language": "system",
  50   │           "name": "shfmt",
  51   │           "pass_filenames": true,
  52   │           "types": [
  53   │             "shell"
  54   │           ]
  55   │         }
  56   │       ],
  57   │       "repo": "local"
  58   │     }
  59   │   ]
  60   │ }
```
</details>

blaggacao:
Merge, 2, 3?
1, 2, merge?

:wink: Thank you @roberth for your guidance!

blaggacao:
```
pre-commit uninstalled
pre-commit installed at .git/hooks/pre-push
```

... now let's call it working!

The code before the last commit only ever installed 'commit' stage.

The code after the commit installs all known stages if they are configured or falls back to 'commit'.
It also removes old, eventually outdated stages.

modules/pre-commit.nix
```diff
@@ -315,7 +329,31 @@ in
...
               else
                 ln -s ${configFile} .pre-commit-config.yaml
-                pre-commit install
+                # Remove any previously installed hooks (since pre-commit iteslf has no idempotent design)
+                hooks="pre-commit pre-merge-commit pre-push prepare-commit-msg commit-msg post-checkout post-commit"
+                for hook in $hooks; do
+                  pre-commit uninstall -t $hook
+                done
+                # Add hooks for configured stages (only) ...
+                if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
```
blaggacao, line 338: Can you confirm `[]` would evaluate to `""`?

Reason: I did not know how to verify. I failed at getting `lib` into the `nix repl`. :confused:

roberth: Confirmed.

modules/pre-commit.nix
```diff
@@ -315,7 +329,31 @@ in
                 echo 1>&2 "    3. add .pre-commit-config.yaml to .gitignore"
               else
                 ln -s ${configFile} .pre-commit-config.yaml
-                pre-commit install
+                # Remove any previously installed hooks (since pre-commit iteslf has no idempotent design)
```
roberth, line 332:
```suggestion
                # Remove any previously installed hooks (since pre-commit itself has no convergent design)
```

modules/pre-commit.nix
```diff
@@ -315,7 +329,31 @@ in
...
+                hooks="pre-commit pre-merge-commit pre-push prepare-commit-msg commit-msg post-checkout post-commit"
+                for hook in $hooks; do
+                  pre-commit uninstall -t $hook
+                done
```
roberth, line 336: This is producing errors like this with the bundled version of pre-commit

```
pre-commit uninstalled
usage: pre-commit uninstall [-h] [--color {auto,always,never}] [-c CONFIG]
                            [-t {pre-commit,pre-merge-commit,pre-push,prepare-commit-msg,commit-msg}]
pre-commit uninstall: error: argument -t/--hook-type: invalid choice: 'post-checkout' (choose from 'pre-commit', 'pre-merge-commit', 'pre-push', 'prepare-commit-msg', 'commit-msg')
usage: pre-commit uninstall [-h] [--color {auto,always,never}] [-c CONFIG]
                            [-t {pre-commit,pre-merge-commit,pre-push,prepare-commit-msg,commit-msg}]
pre-commit uninstall: error: argument -t/--hook-type: invalid choice: 'post-commit' (choose from 'pre-commit', 'pre-merge-commit', 'pre-push', 'prepare-commit-msg', 'commit-msg')
pre-commit installed at .git/hooks/pre-commit
```

I don't see a a reason not to uninstall pre-commit entirely before installing. It's also more future proof.

modules/pre-commit.nix
```diff
@@ -315,7 +329,31 @@ in
...
+                  for stage in ${concatStringsSep " " cfg.default_stages}; do
+                    case $stage in
+                      commit | merge-commit | push)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
+                      prepare-commit-msg | commit-msg | post-checkout | post-commit)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
```
roberth, line 348: I see what you're trying to do here, but it's broken. Let's stick closely to git and pre-commit's terminology. It's easier to learn for those who are familiar with it and it's more future proof (any new hooks will be ignored)

modules/pre-commit.nix
blaggacao: I _think_ `pre-commit` doesn't behave as we wised it to behave, see:

```console
$ pre-commit install
pre-commit installed at .git/hooks/pre-commit

$ pre-commit install -t post-checkout
pre-commit installed at .git/hooks/post-checkout

$ pre-commit uninstall
pre-commit uninstalled

# post-checkout is still installed
$ ll .git/hooks/post-checkout
.rwxrwxr-x blaggacao blaggacao 1.4 KB Tue Jul 21 12:02:26 2020  post-checkout

# Only now it's cleaned up
$ pre-commit uninstall -t post-checkout
post-checkout uninstalled
```
The solution is to either bundle another pre-commit or take out `post_*` hooks from the list

modules/pre-commit.nix
blaggacao: We have to acknowledge it's a fundamental flaw of pre-commit:

1. On one hand it tries to be as declarative as possible in version controlled `.pre-commit-config.yml`
2. On the other hand any _sound_ distributed configuration on the developer machine is out of the reach of the devops team (without employing an auxiliary task runner, such as `make` or `just`).

Since a `shell.nix` is expected to produce a reproducible environment, this flaw is fundamentally incompatible with `shell.nix` philosophy.

Hence, since we don't want to betray ourselves (`shell.nix`), fixing this non-converging design of pre-commit is a moral imperative.

The side effects introduced by pre-commit into the local `.git/hooks/...` _must_ be controlled.

I'd rather prefer to do help message parsing to make this future proof, than to let those side effects rot in there to eternity.

blaggacao: I filed an upstream feature request: https://github.com/pre-commit/pre-commit/issues/1541

blaggacao:
@roberth 
I'd like to apologize for my imperative opinions. However, without controlling the side effects, my use case gets completely jeopardized..

I'd suggest for the intermediate term to adopt a work around with a more involved error catcher.

Let's track https://github.com/pre-commit/pre-commit/issues/1541 in complement to this course of action in a separate issue.

Would that be in line with your concerns?

blaggacao:
We maybe, more fundamentally, have to ask ourselves, what stance to take on the pre-commit author's intentions:

> pre-commit intentionally makes opting in and out an explicit action

Since in the current implementation `pre-commit install` (with or without this PR) is triggered, this would violate the author's intention.

O the other hand, shell.nix holds promises about repeatable _development" environments. Opting out is still possible.

domenkozar:
Since the tool is called `pre-commit` I would expect the default to be running hooks before committing.

blaggacao:
> Since the tool is called `pre-commit` I would expect the default to be running hooks before committing.

Fair enough :smile: 

In reality though, pre-commit is evolving into a _repo-scoped_ (per-repo cache) packager for task runners of all sorts.

To that end, I really like that you cut this sprout and called it `local` :+1: 

So once we strip that we are left with — despite the name — a _git_ hook manager similar to [`husky`](https://github.com/typicode/husky#readme).

Husky, contrary to pre-commit [_does_ make version controlled pretentions](https://github.com/typicode/husky#install) about which hooks to be installed.

See also:
> `npm uninstall husky`
> Git hooks installed by husky will be removed.

I like husky's approach better and I think we should take the liberty and "fix" pre-commit for the nix community.

I can assure you, pre-commit's model is not useful as-is for me who has to make sure my collaborator is always up to date with the latest coding guidelines. That's wy I'm turning to nix altogether in the first place.

modules/pre-commit.nix
roberth: > moral imperative

I'm taking this with a shovel of salt, considering that it's just some tech in the end. I wouldn't be so sure if someone outside the functional programming bubble (like pre-commit's maintainer and the python community) appreciates this kind of 'grand' language about concepts that are unnecessary from a pragmatic point of view.

Anyway, this seems to be about the broader topic of improving declarative hook installation, whereas my comment was originally intended about the `"pre-"` prefixing. I totally failed to make that clear. The idea is that we should treat the names of commit hooks atomically rather than making them the subject of string manipulation. The reason to do so is that it'd be easier for users to switch between pre-commit-hooks.nix configuration, git manpages and pre-commit.com documentation when the terminology is aligned one-to-one.

modules/pre-commit.nix
roberth: Updating our pins here is totally fine. We use `niv` for that.

roberth:
At first glance, husky doesn't seem to provide help with the optimization of running only on modified files. Also iirc some users of this project also use or want to use existing non-Nixified pre-commit "repos" (packaged via other language-specific means)

I don't think it would be beneficial to kick out pre-commit in favor of another tool for these reasons.
The issue this PR is facing is minor. The lack of a `pre-commit uninstall --all` can be worked around, as you have done. Sure that means maintaining a list of hooks, but that's far better than fragmenting the Nix community's git hook packaging effort.

blaggacao:
@roberth Totally agree with both apsects (keeping pre-commit and working around its limitations). 

Thanks for your opinion!

Counting your vote, my implicit vote (via this PR) and @domenkozar apparent abstention for a clear vote (please correct), I might summarize that we agree moving forward with what this PR tries to attempt.

I have already deployed it in our organization and did a last fixup yesterday.

Therefore, may I suggest you proceed with a formal incremental review?

blaggacao:
Oh, I'll have a look at the tests after lunch.

modules/pre-commit.nix
```diff
@@ -59,7 +59,7 @@ let
                       The entry point - the executable to run. entry can also contain arguments that will not be overridden such as entry: autopep8 -i.
                     '';
                 };
-              language =
+             language =
```
roberth, line 62:
```suggestion
              language =
```

modules/pre-commit.nix
roberth: Is there any reason not to simplify the whole `case ... esac` statement into just `pre-commit install -t $stage`? The second clause is equal to the first (probably by mistake?) but I don't think we really need to do anything special here anyway.

blaggacao: by mistake, thanks!

In `pre-commit`, stages `commit | merge-commit | push` are named different from hooks `pre-commit | pre-merge-commit | pre-push`. It's very confusing, but who are we to tell... :wink:

blaggacao: force pushed

roberth: Ok, we'll just deal with that. No problem. One more comment about the default case though, but I'll make that a separate thread.

modules/pre-commit.nix
blaggacao: This is getting taxing... :wink:
```console
$ niv add pre-commit
Adding package pre-commit
  FATAL: One or more packages failed to update:
  pre-commit: Update failed: Key could not be found: url_template
  CallStack (from HasCallStack):
    error, called at src/Niv/Update.hs:80:25 in niv-0.2.13-4DFkDQryuRMJTtP7ftbwS:Niv.Update
ERROR: ExitFailure 1
```

modules/pre-commit.nix
```diff
@@ -315,7 +329,43 @@ in
...
+                    case $stage in
+                      commit | merge-commit | push)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
+                      prepare-commit-msg | commit-msg | post-checkout | post-commit)
+                        pre-commit install -t $stage
+                        ;;
+                      *)
+                        ;;
```
roberth, line 362: iiuc at this point we're dealing with a `$stage` that's probably valid, according to the check with `$install_help =~ $stage`. Instead of doing nothing and failing to install silently, we should allow `pre-commit install -t $stage` to try, or fail with a message that the `$stage` hook is not supported yet.

I think that's the only remaining issue now. Thanks for being patient :)

modules/pre-commit.nix
blaggacao: > Updating our pins here is totally fine. We use niv for that.

Oh I had overlooked. Could you check if the last commit gets it right? (I'm new to niv).

modules/pre-commit.nix
blaggacao: > Thanks for being patient

No worries! I'm aware this is done once and used many.

Since in my last commit I removed the _shady_ regex on the help command, I think you are right.

I'll do the following:
- keep the switch  `prepare-commit-msg | commit-msg | post-checkout | post-commit)` - since we only can uninstall so many (they have to be in $hook`)
- add an error message to `*)`

roberth:
> EDIT: @domenkozar @roberth What do you think of out-factoring the install phase into another optional option, so we more faithfully maintain the opt-in + opt-out nature of pre-commit?

You could attach the shellHook to something that isn't the projects `/shell.nix`

blaggacao:
> You could attach the shellHook to something that isn't the projects /shell.nix

Nice! Freedom of the user recovered! Peace! :waffle:

modules/pre-commit.nix
blaggacao: I'll furthermore exit with 1, since we should care about expectations matching the environment.

blaggacao:
https://github.com/cachix/pre-commit-hooks.nix/pull/52/commits/2556038aa1799c08659a1ed47e978bf191774086 looks green but I have no idea if I'm doing the right thing. _darknix_ again :stuck_out_tongue_winking_eye: .

modules/pre-commit.nix
roberth: Looks kinda alright, although I don't know why yours has a `version` attribute. I get no `version` when running

    nix-shell
    niv drop pre-commit
    niv add pre-commit/pre-commit

Did you use the nix-shell in the repo?

blaggacao: [`niv add pre-commit/pre-commit -v v2.6.0`](https://github.com/pre-commit/pre-commit/releases/tag/v2.6.0)

I don't think there are guarantees that `pre-commit` master works all the time.

blaggacao:
Now that we've got the design mostly right, let's do a nitpicking round?

modules/pre-commit.nix
roberth: Oh no, seems like `-v` doesn't do anything. https://github.com/nmattia/niv/issues/232
I normally use `-b TAG` which works, even though it was named for branches.

modules/pre-commit.nix
```diff
@@ -315,7 +329,36 @@ in
...
+                      # if you amend these switches please also review $hooks above
+                      commit | merge-commit | push)
+                        stage="pre-"$stage
+                        pre-commit install -t $stage
+                        ;;
+                      prepare-commit-msg | commit-msg | post-checkout | post-commit)
+                        pre-commit install -t $stage
+                        ;;
+                      *)
+                        echo 1>&2 "ERROR: nix-pre-commit-hooks: either $stage is not a valid stage or pre-commit-hook.nix doesn't yet support it."
```
roberth, line 353:
```suggestion
                        echo 1>&2 "ERROR: nix-pre-commit-hooks: either $stage is not a valid stage or pre-commit-hooks.nix doesn't yet support it."
```

modules/pre-commit.nix
blaggacao: Another _darknix_ caveat :smirk:

roberth:
The code assumes `default_stages` are exactly the stages that should be installed. This won't be the case when support for per-hook `stages` is added. We can choose to solve that later though; doesn't seem breaking.

blaggacao:
> The code assumes default_stages are exactly the stages that should be installed. This won't be the case when support for per-hook stages is added. We can choose to solve that later though; doesn't seem breaking.

Yep, I thought about that, too. It should be rather straight forward to add that here:
```
     if [ ! -z "${concatStringsSep " " cfg.default_stages}" ]; then
                  for stage in ${concatStringsSep " " cfg.default_stages}; do

```

blaggacao:
> The code assumes `default_stages` are exactly the stages that should be installed. This won't be the case when support for per-hook `stages` is added. We can choose to solve that later though; doesn't seem breaking.

I decided adding a hint to future implementer.

nix/sources.json
```diff
@@ -46,5 +46,17 @@
         "type": "tarball",
         "url": "https://github.com/NixOS/nixpkgs/archive/99a3d7a86fce9e9c9f23b3e304d7d2b1270a12b8.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "pre-commit": {
```
domenkozar, line 50: Why is this here?

blaggacao: Intent: get lates version of pre-commit which also supporte post-* hooks. (The previously used version didn't.) Pinning to a well known version made the converging install/uninstall bash code less brittle.

Why is it there: due to @roberth 's suggestion as to the means to achieve said intent.

blaggacao: Details https://github.com/cachix/pre-commit-hooks.nix/commit/2556038aa1799c08659a1ed47e978bf191774086

blaggacao:
Shall we proceed?

nix/sources.json
domenkozar: It doesn't seem to me this is used though? `cfg.package` will still use `pkgs.pre-commit`.

Maybe it we should bump `nixpkgs` to master instead.

blaggacao: Oh, ok. I learnd something. Thanks, will do.

blaggacao: Done

blaggacao:
Ahm... From the CI.
```console
 error: attribute 'Cabal_3_0_0_0' missing, at /home/runner/work/pre-commit-hooks.nix/pre-commit-hooks.nix/nix/default.nix:20:27
(use '--show-trace' to show detailed location information)
##[error]Process completed with exit code 1.
```

I hope that's not because of my latest commit, if so, I'd ask you to solve this the right way so I can learn by example.

blaggacao:
@domenkozar What's holding this back? Is it time?

domenkozar:
Bumping nixpkgs to master broke a few things, so we probably best revert that and have two versions of nixpkgs for tooling. I think eventually we'll need that sooner than later.

blaggacao:
@domenkozar Ok I reverted it but I do unfortunately not know how to enable two `nixpkgs` versions.

domenkozar:
Since I don't have time for this I recommend we wait for 20.10 to come out and pin that one which should also unblock this merge.

domenkozar:
I've bump nixpkgs in #65 and this issue is a good reminder that #61 is quite error prone.

domenkozar:
Alright master is ready!

blaggacao:
replaced by https://github.com/cachix/pre-commit-hooks.nix/pull/79

PR #53, opened 2020-07-20
Suggest default_stages

blaggacao:
Follow up to #52

I like this config.

Speedy people on speedy command lines on speed might like it, too.

I hate pre-commit nagging on every marginal commit or amend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ The goal is to **manage commit hooks with Nix** and solve the following:
     in {
       pre-commit-check = nix-pre-commit-hooks.run {
         src = ./.;
-        # Leaving out this option falls back to install all hooks into the 'commit' stage
-        default_stages = ["manual" "push"];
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,6 +29,8 @@ The goal is to **manage commit hooks with Nix** and solve the following:
     in {
       pre-commit-check = nix-pre-commit-hooks.run {
         src = ./.;
+        # Leaving out this option falls back to install all hooks into the 'commit' stage
+        default_stages = ["manual" "push"];
```
roberth, line 33: We should stick to pre-commit defaults by default in our examples, to minimize surprises. I don't think we should suggest this to novices if they're only running some quick formatters.

```suggestion
        # If your hooks are intrusive, avoid running on each commit with a default_states like this:
        # default_stages = ["manual" "push"];
```

Also avoids a negation :)

domenkozar: I agree with Robert, we shouldn't modify the default of pre-commit project.

PR #58, opened 2020-08-24
Add Black for Python

akirak:
This is a formatter for Python.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -36,6 +36,13 @@ in
             "Ansible linter";
           entry = "${tools.ansible-lint}/bin/ansible-lint";
         };
+      black =
+        {
+          name = "black";
+          description = "The uncompromising Python code formatter";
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -36,6 +36,13 @@ in
             "Ansible linter";
           entry = "${tools.ansible-lint}/bin/ansible-lint";
         };
+      black =
+        {
+          name = "black";
+          description = "The uncompromising Python code formatter";
+          entry = "${pkgs.python3Packages.black}/bin/black";
+          files = "\\.py$";
```
roberth, line 44: It's better to use the `python` [`type`](https://pre-commit.com/#filtering-files-with-types) which will also find scripts with python shebangs. (and potentially more?)

akirak: I didn't know of `types`. Thanks, fixed.

akirak:
Fixed the issue.

Issue #35, opened 2020-02-06
Shellcheck: files are not picked up correctly


roberth:
`types: bash` only finds `#!/bin/bash` shebangs it seems.

It should pick up:
 - `*.sh`
 - `#!/bin/bash` (this one seems to work)
 - `#!/bin/sh`
 - `#!/usr/bin/env bash`
 - `#!/usr/bin/env sh`
 - `#!/usr/bin/env nix-shell`+`#!.*nix-shell.*-i bash.*`

domenkozar:
This should probably be reported to https://github.com/pre-commit/pre-commit, since filtering isn't advanced enough to cover all these cases.

blaggacao:
Not even there, but here: https://github.com/chriskuehl/identify

blaggacao:
Please review & approve there: https://github.com/chriskuehl/identify/pull/127

blaggacao:
After those changes, `#!.*nix-shell.*-i bash.*` this unconventionak format would not be supported by `identify`, yet. But maybe ther is an argument for supporting it cleanly to be made upstream.

blaggacao:
Yoou! Looks like striking this ticket off the list comes within reach :partying_face:

asottile:
OP's message implies this is using `types: [bash]` -- you probably want `types: [shell]` fwiw

blaggacao:
@asottile 
> OP's message implies this is using types: [bash] -- you probably want types: [shell] fwiw

Oh thank you! Actually the following commit was meant not to hit on eg. `csh` which is not supported by shellcheck. I can already update the corresponding draft PR here. Is there a possibility to cut a release for `identify` soonish?

https://github.com/chriskuehl/identify/pull/129

asottile:
I was going to merge the nix stuff first

roberth:
We can use this list to set the right `types` https://github.com/koalaman/shellcheck/blob/cfd68ee0c2ebfd0ab08a1d4bf628162b454dc207/src/ShellCheck/Parser.hs#L3236

roberth:
Encountered this just now. Not sure if everything is up to date, but I suppose this may help others. My error:
```
In scripts/generate-nix line 1:
#!/usr/bin/env nix-shell
^-- SC1008: This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.

```

Successful workaround:

```
#!/usr/bin/env nix-shell
[...]

PR #63, opened 2020-09-04
towards #35

blaggacao:
#35

---

**Do not merge before** https://github.com/chriskuehl/identify/pull/129 landed in a release which is then _acctually_ used here.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -128,6 +128,7 @@ in
           name = "shellcheck";
           description = "Format shell files";
           types =
+            # based on `goodShells` in https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Parser.hs
             [
               "sh"
               "ash"
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -129,7 +129,7 @@ in
           description = "Format shell files";
           types =
             [
-              "bash"
+              "shell"
             ];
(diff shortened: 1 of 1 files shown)


roberth:
I don't think we want to check unsupported shells though. https://github.com/koalaman/shellcheck/blob/cfd68ee0c2ebfd0ab08a1d4bf628162b454dc207/src/ShellCheck/Parser.hs#L3236

blaggacao:
```hs
    goodShells = [
        "sh",
        "ash",
        "dash",
        "bash",
        "bats",
        "ksh"
        ]
```

blaggacao:
The problem will be `*.sh` extension since I believe `identity` gives precedence to extension over shebang. Not sure though.

blaggacao:
**Do not merge before** https://github.com/chriskuehl/identify/pull/129 landed in a release which is then _acctually_ used here.

modules/hooks.nix
```diff
@@ -129,7 +129,12 @@ in
           description = "Format shell files";
           types =
             [
```
roberth, line 131:
```suggestion
            # based on `goodShells` in https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Parser.hs
            [
```

domenkozar:
identify now just needs a bump in nixpkgs

domenkozar:
Going to fix as part of #65 that will include the bump.

Issue #61, opened 2020-09-04
Make usable without package distribution


roberth:
pre-commit-hooks.nix in its current form is a package distribution by default; using pinned packages.

Some users (me, #55?) prefer to provide their own packages only, in order to have consistent versions everywhere in a project.

I'd be happy to provide a function to do that, which will work at the module level.

TODO
 - [x] write #60 to make the module system interface a similar as possible
 - [ ] use module system defaults for the default tools
 - [ ] set the default tools with a module that can be enabled (and enable it in `run` so its behavior remains unchanged)
 - [ ] expose the function in `default.nix`

In order to improve performance I'd like to move `tools` into its own attrset, so we can evaluate the spine of `default.nix` without having to load the pinned nixpkgs. It's even possible to unbreak the move and provide deprecation warnings.

blaggacao:
If &mdash; as a collateral &mdash; the code base could be made easier to grasp, that would be great. I was in sheer despair more than once when working on #52 . &mdash; not directly related, though.

domenkozar:
Couldn't we just expose `pkgs` attribute that could be overriden?

roberth:
> Couldn't we just expose `pkgs` attribute that could be overriden?

In the past we've needed a solution for packages that are packaged with `pre-commit-hooks.nix` itself. Those did not come from `pkgs` so to support that, we'd need to expose a way to override `tools` (or something similar).

So either we forgo packaging hooks or the ability to override those, or we allow `tools` overrides. The latter can already be done if you wire up the modules yourself, so I think we should make `run` do nothing more than invoke the module system. Not rearranging arguments into a configuration module is a small breaking change iirc; not too bad.

domenkozar:
Overriding tools sounds good to me (I forgot we didn't just apply an overlay to pkgs).

blaggacao:
I like the way how devshell presents itself as an overlay, this makes it easier to chain things together. Maybe not 100% on topic, but I figured to best comment here nevertheless: https://github.com/numtide/devshell/blob/master/default.nix

u-quark:
[...]

Issue #55, opened 2020-07-24
Ability to select specific versions of tooling


fiadliel:
Whether through passing in your own nixpkgs, sources.json, or otherwise, it would be nice to be able to influence the versions of tooling chosen, without having to fork the project.

A concrete case:
haskell-language-server for VSCode uses a specific version of ormolu for formatting, and this is built-in to the server as a library, not binary (with current hls, 0.1.2.0). If we can't align ormolu with the version pre-commit-hooks uses, then the feature in HLS/VSCode cannot reasonably be used concurrently with hooks (which is a pity for usability etc.).

domenkozar:
It's possible to do that via `pre-commit.tools` or `entry` but you have to read the source to get it right.

It would be far better if we exposed package per each tool that could be overriden.

domenkozar:
Similar to #61

domenkozar:
This is now fixed, you can set `package` for each hook!

PR #67, opened 2020-09-24
Add nix flakes support

### LINKED ISSUE #57: Expose as a flake
Currently its not possible to use this with `nix develop` because it internally re-imports nixpkgs:

```
error: --- EvalError --------------------------------------------------------------------------------- nix
at: (90:39) in file: /nix/store/ma3b77vjls3qrznwh66ix5ivvybx6003-source/pkgs/top-level/impure.nix

    89|                 else (if args ? localSystem then {}
    90|                       else { system = builtins.currentSystem; }) // localSystem;
      |                                       ^
    91| })

attribute 'currentSystem' missing
(use '--show-trace' to show detailed location information)
direnv: export +IN_NIX_SHELL ~PATH
```

Myhlamaeus:
This adds a `flake.nix` which exposes the project's outputs (mainly `packages.{system}.run`) and defines the dependencies.

`niv` was removed as nix flakes obviate it.

BREAKING CHANGES: The project.nix module is now exposed as an output
`projectModules.pre-commit-hooks` on the flake instead of a separate file.

Fixes #57


Myhlamaeus:
I'll soon create a corresponding PR on hercules-ci/project.nix as this PR breaks compatibility with it.

domenkozar:
This needs a rebase on top of master that landed 20.09 and other things :)

Stupremee:
Is this planned to be merged any time soon?

roberth:
We should only offer a cleaned-up interface in the flake. Otherwise people will have to migrate twice - once for flakes and then again for the revised interface. This means implementing some of the ideas in #55 and #61.
This also means we can get rid of some of the inputs and take them from Nixpkgs.
If you enjoy removing unnecessary wiring, I'd be happy to review. Otherwise I can do it, but probably no earlier than a month from now.

Stupremee:
I don't think I have enough experience with Nix to do it.

terlar:
Did this result in anything, I see nothing linked here, so I guess not, but just checking as this would be a nice feature to have.

dhess:
Even just a simple fix for the `currentSystem` error when evaluating this project in a flake would be really helpful:

```
error: attribute 'currentSystem' missing

       at /nix/store/smvan4qznfgg79w1s6wjw1zdbszkynav-source/pkgs/top-level/impure.nix:90:39:

           89|                 else (if args ? localSystem then {}
           90|                       else { system = builtins.currentSystem; }) // localSystem;
             |                                       ^
           91| })
(use '--show-trace' to show detailed location information)
```

domenkozar:
> Even just a simple fix for the `currentSystem` error when evaluating this project in a flake would be really helpful:
> 
> ```
> error: attribute 'currentSystem' missing
> 
>        at /nix/store/smvan4qznfgg79w1s6wjw1zdbszkynav-source/pkgs/top-level/impure.nix:90:39:
> 
>            89|                 else (if args ? localSystem then {}
>            90|                       else { system = builtins.currentSystem; }) // localSystem;
>              |                                       ^
>            91| })
> (use '--show-trace' to show detailed location information)
> ```

I'm happy to fix this - what's the easiest way to reproduce it?

dhess:
@domenkozar Thanks for the offer! This public project has a `source-code-checks` attribute which builds fine using `nix-build` but not using `nix build` (with flakes support enabled):

https://github.com/hackworthltd/haskell-template/blob/a51b92252120a7fe82bc4d9f70842aae3526cbe8/flake.nix#L82

Beware that it uses `haskell.nix` so you'll probably want to be using the upstream `haskell.nix` binary cache. Alternatively, you could just take a look at what I'm doing there (it's pretty straightforward) and add it to a simple test flake.

BTW, I thought this was going to be a very easy fix as it looks like you just need to set `localSystem.system` to `system` in `nixpkgs`s args, but that still didn't work. If I recall, it has something to do with `niv`.

domenkozar:
Sounds like the best way is to add flake support into the project itself.

I'm going to close this and track progress at https://github.com/cachix/pre-commit-hooks.nix/issues/57

domenkozar:
Thanks @Myhlamaeus for your hard work, I've incorporated some of your changes and flake support is live now: https://github.com/cachix/pre-commit-hooks.nix#nix-flakes-support

PR #72, opened 2020-10-09
Run hlint in --refactor mode (inplace if possible)

### LINKED ISSUE #69: Run hlint in --refactor mode
Is it possible to add this as the project it is right now? Just like `ormolu` is run with `--mode inplace`? 😉

kutyel:
Fixes #69 

1) Maybe I need to add https://github.com/mpickering/apply-refact somewhere? 
2) Maybe we want to somehow opt-out of the "inplace" mode in the case of hlint 🤔
3) Is it possible to add the label `hacktoberfest-accepted` to this PR? 🙏 

Happy hacktoberfest! 🥳

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -53,7 +53,7 @@ in
       hlint =
         {
           name = "hlint";
-          description = 
+          description =
             "HLint gives suggestions on how to improve your source code.";
           entry = "${tools.hlint}/bin/hlint --refactor --refactor-options='--inplace'";
           files = "\\.l?hs$";
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -53,9 +53,9 @@ in
       hlint =
         {
           name = "hlint";
-          description =
+          description = 
             "HLint gives suggestions on how to improve your source code.";
-          entry = "${tools.hlint}/bin/hlint";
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -53,9 +53,9 @@ in
       hlint =
         {
           name = "hlint";
-          description =
+          description =
```
domenkozar, line 56:
```suggestion
          description =
```

dhess:
This doesn't work with `hlint-3.2.1`, at least:

```
hlint....................................................................Failed
- hook id: hlint
- exit code: 1

hlint: Refactor flag can only be used with an individual file
CallStack (from HasCallStack):
  errorIO, called at src/HLint.hs:194:14 in hlint-3.2.1-5SE9YOnH4UyLLY7L7KPrCd:HLint
```

domenkozar:
That's strange, I think pre-commit does pass each file as an argument.

dhess:
Yes, this is a fully checked-in & committed source tree.

domenkozar:
I've reverted this as it seems that hlint only supports this option iff one file is passed: https://github.com/ndmitchell/hlint/blob/master/src/HLint.hs#L194

PR #77, opened 2020-10-31
nixpkgs-fmt: include ".ignore" file

gilligan:
This changes the default of `nixpkgs.files` to include an `.ignore`
file which nixpkgs-fmt reads to register exclusions.

Without this change, the behavior between running nixpkgs-fmt inside and
outside of the pre-commit hook is not consistent.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -110,7 +110,7 @@ in
           name = "nixpkgs-fmt";
           description = "Nix code prettifier.";
           entry = "${tools.nixpkgs-fmt}/bin/nixpkgs-fmt";
-          files = "\\.nix$";
+          files = "(\\.nix$)|(\\.ignore$)";
         };
       nix-linter =
         {


domenkozar:
Ugh, I wish the name wouldn't be so generic :)

gilligan:
> Ugh, I wish the name wouldn't be so generic :)

I agree, but that would be an upstream issue. For now i would like to get a consistent behavior which currently is only possible with manually maintaining a .ignore file and excludes for the hook, which is really bad.

domenkozar:
It's a bit surprising this is needed, can't nixpkgs-fmt just read that file? I'm a bit surprised it has to be passed as a file to be formatted.

gilligan:
Well this is the *only* way i got it to work. You can try it yourself. Create an .ignore file and add some pattern to it. Change the excluded file and run nixpkgs-fmt. It will ignore it. Now run git commit: the hook will fail with nixpkgs-fmt wanting to format that file.

If you find a different solution to this: fine by me of course 😅

domenkozar:
I will merge this, just need to understand how it works cc @zimbatm

gilligan:
I am happy to admit that i don’t understand this either. I thought maybe some filtering is going on making the ignore file invisible to the hook ?! It’s a bit weird.

zimbatm:
nixpkgs-fmt inherits that behaviour from the https://crates.io/crates/ignore crate, that is being used to walk the filesystem.

It supports .gitignore and .ignore files by default, and should be probably extended with its own filename.

Since pre-commit-hook is passing the files directly, I don't know what the right behaviour should be. It looks like `ignore` opted to only take `.ignore` files into account when passed explicitly which seems a bit weird to me.

domenkozar:
@gilligan OK, if you add a comment why ignore file is needed (summarizing our discussion), let's merge this.

gilligan:
@domenkozar actually i think this is still broken. This is really annoying..

PR #79, opened 2020-11-02
feat: support specifying stages, close #49

### LINKED ISSUE #49: Support specifying stages

blaggacao:
replace #52 (rebased & squashed)

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -277,6 +277,7 @@ in
               A configuration wide option for the stages property.
               Installs hooks to the defined stages.
               Default is empty which falls back to 'commit'.
+              See https://pre-commit.com/#confining-hooks-to-run-at-certain-stages
             '';
           default = [ ];
         };
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -15,6 +15,7 @@ let
   inherit (pkgs) runCommand writeText git;
 
   cfg = config;
+  install_stages = cfg.default_stages;
 
   hookType =
     types.submodule (
@@ -268,6 +269,18 @@ in
           default = [ ];
         };
 
+      default_stages =
+        mkOption {
+          type = types.listOf types.str;
+          description =
+            ''
+              A configuration wide option for the stages property.
+              Installs hooks to the defined stages.
+              Default is empty which falls back to 'commit'.
+            '';
+          default = [ ];
+        };
+
       rawConfig =
         mkOption {
           type = types.attrs;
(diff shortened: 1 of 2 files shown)


blaggacao:
beware I'm hitting this:

```console
$ pre-commit run --all --hook-stage manual generate
An error has occurred: InvalidConfigError:
==> File .pre-commit-config.yaml
==> At Config()
==> At key: repos
==> At Repository(repo='local')
==> At key: hooks
==> At Hook(id='shellcheck')
==> At key: types
==> At index 1
=====> Type tag 'ash' is not recognized.  Try upgrading identify and pre-commit?
Check the log at /home/blaggacao/.cache/pre-commit/pre-commit.log
$ pre-commit --version
pre-commit 2.5.1
```

modules/pre-commit.nix
```diff
@@ -325,7 +340,36 @@ in
...
+                done
+                # Add hooks for configured stages (only) ...
+                if [ ! -z "${concatStringsSep " " install_stages}" ]; then
+                  for stage in ${concatStringsSep " " install_stages}; do
+                    if [[ "$stage" == "manual" ]]; then
+                      continue
+                    fi
+                    case $stage in
+                      # if you amend these switches please also review $hooks above
+                      commit | merge-commit | push)
```
domenkozar, line 356: Why do we diverge from pre-commit settings by not just specifying pre-push instead of push?

modules/pre-commit.nix
```diff
@@ -268,6 +269,18 @@ in
...
         };
 
+      default_stages =
+        mkOption {
+          type = types.listOf types.str;
+          description =
+            ''
+              A configuration wide option for the stages property.
+              Installs hooks to the defined stages.
+              Default is empty which falls back to 'commit'.
```
domenkozar, line 279:
```suggestion
              Default is empty which falls back to 'commit'.
              See https://pre-commit.com/#confining-hooks-to-run-at-certain-stages
```

modules/pre-commit.nix
blaggacao: Precommit has two identifiers "push" (the stage) and "pre-push" (the hook) in an attempt to be extra "correct" about things, while achieving only plain confusion. Here it's "push" (the stage).

blaggacao: Ps: I also wondered and got bitten. More than once.

PR #87, opened 2020-12-02
Accept nixpkgs and system as arguments

gilligan:
This allows passing in of `nixpkgs` and `system` attributes to
default.nix and nix/default.nix

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,1 +1,4 @@
-let pkgs = import ./nix { }; in pkgs.packages
+let
+  pkgs = import ./nix { };
+in
+pkgs.packages
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,5 +1,6 @@
 { sources ? import ./sources.nix
 , system ? builtins.currentSystem
+, nixpkgs ? sources.nixpkgs
 }:
 let
   overlay =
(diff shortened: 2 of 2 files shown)


gilligan:
@domenkozar any strong feelings about this? With the default arguments in place this shouldn't have any effect on users while allowing to pass in nixpkgs/system when wanted. I'm happy to change any implementation details according to your preference of course though.

gilligan:
Actually i am an idiot and this is indeed a breaking change because default.nix now needs to be called.

Just let me know what you think @domenkozar

domenkozar:
I'm afraid this will open pandora's box of having glue code compatibility of different nixpkgs versions, but let's see how much mess there is.

domenkozar:
I wonder if you can just import `nix/default.nix` instead of breaking backwards compatibility?

gilligan:
> I wonder if you can just import nix/default.nix instead of breaking backwards compatibility?

yeah.. let's not annoy users :sweat_smile: I updated the PR to that effect. WDYT?

default.nix
```diff
@@ -1 +1,7 @@
-let pkgs = import ./nix { }; in pkgs.packages
+let
+  sources = import ./nix/sources.nix;
+  pkgs = sources.nixpkgs;
+  system = builtins.currentSystem;
+  _pkgs = import ./nix { inherit sources system; nixpkgs = pkgs; };
+in
+_pkgs.packages
```
roberth, line 7: Why duplicate the wiring? These are all default arguments.

gilligan: That is.... a good question and the answer is that i just amended my commit. Thanks @roberth

roberth (approved):

PR #89, opened 2020-12-27
Add Fourmolu (Haskell formatter)

georgefst:
[Fourmolu](https://github.com/parsonsmatt/fourmolu) is a fork of Ormolu, which allows configuration and has quite different default behaviour.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -87,6 +87,16 @@ in
             }";
           files = "\\.l?hs$";
         };
+      fourmolu =
+        {
+          name = "fourmolu";
+          description = "Haskell code prettifier.";
+          entry =
+            "${tools.fourmolu}/bin/fourmolu --mode inplace ${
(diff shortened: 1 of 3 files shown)


modules/hooks.nix
```diff
@@ -87,6 +87,16 @@ in
             }";
           files = "\\.l?hs$";
         };
+      fourmolu =
+        {
+          name = "fourmolu";
+          description = "Haskell code prettifier.";
+          entry =
+            "${tools.fourmolu}/bin/fourmolu --mode inplace ${
+            lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) settings.ormolu.defaultExtensions)
```
domenkozar, line 96: Shouldn't settings come from `fourmolu` as well?

georgefst: Well, we track upstream very closely. The CLI is unlikely to ever diverge.

We could add a `settings.fourmolu` instead with the exact same implementation, but I think I prefer it this way.

PR #92, opened 2020-12-30
Add config option to add custom user defined hooks

### LINKED ISSUE #86: Implementing an ad-hoc pre-commit hook
The way I was thinking about this previously was my naive understanding of nix. I ended up forking the repository and adding my custom hook but I am still curious if one can implement an ad-hoc pre-commit hook without forking the entire repo.

**edit**: Ignore anything below this line

I see https://github.com/cachix/pre-commit-hooks.nix#contributing-hooks in the documentation but I think it would be helpful to also document how to create ad-hoc pre-commit hooks.

Here is my attempt at adding a custom hook: https://github.com/chiroptical/declarative-programming-streams/blob/64dec6e89fbed46a66d08f638d38a5f8d83da942/default.nix#L1-L25, the error:

```sh
$ nix-build
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels/nixpkgs' does not exist, ignoring
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
error: The option `hooks.build-readme.entry' is used but not defined.
(use '--show-trace' to show detailed location information)~~
```

I don't know where to go from here and was hoping someone could give me a little guidance. I would be happy to write the section in the README afterwards on custom hooks.

u-quark:
Sometimes it is useful to add a project specific command as an extra check that is not part of the pre-defined set of hooks provided by this project.

Closes #86.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,59 @@ used at the same time.
 
 - [hunspell](https://github.com/hunspell/hunspell)
 
+## Custom hooks
+
+Sometimes it is useful to add a project specific command as an extra check that
+is not part of the pre-defined set of hooks provided by this project.
+
+Example configuration:
+```nix
+ let
+   nix-pre-commit-hooks = import (builtins.fetchTarball "https://github.com/cachix/pre-commit-hooks.nix/tarball/master");
+ in {
+   pre-commit-check = nix-pre-commit-hooks.run {
+     hooks = {
+       # ...
+
+       # Example custom hook for a C project using Make:
+       unit-tests = {
+         enable = true;
+
+         # The name of the hook (appears on the report table):
+         name = "Unit tests";
+
+         # The command to execute (mandatory):
+         entry = "make check";
+
+         # The pattern of files to run on (default: "" (all))
+         # see also https://pre-commit.com/#hooks-files
+         files = "\\.(c|h)$";
+
+         # List of file types to run on (default: [ "file" ] (all files))
+         # see also https://pre-commit.com/#filtering-files-with-types
+         # You probably only need to specify one of `files` or `types`:
+         types = [ "text" "c" ];
+
+         # Exclude files that were matched by these patterns (default: [ ] (none)):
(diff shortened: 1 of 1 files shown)


roberth (changes requested):
Hi @u-quark, thank you for helping out with custom hooks. They were already supported but not documented though (see suggestion comment). If we change things around a bit we can make it even better.

The reason I think we should avoid the new syntax is because it is inconsistent, creating a difference between custom and built-in hooks, which is not needed. It's similar to NixOS's `users.extraUsers` tech debt. The hidden strength of the module system is that "built-in" modules aren't any different from configuration modules.

We could change the default value of `enable` to `true` and explicitly disable them in the built-in hooks. I don't expect that to break much, because it seems safe to assume that without documentation for it, nobody creates their own 'libraries' of hooks.

u-quark:
This is so much better! I tried doing this before but run into a misleading error and assumed it is not possible. Thanks for the help. I updated this PR to only include documentation for how to do this based on your suggestion.

I don't think the `enable` flag is a big deal at the moment.

u-quark:
Thanks!

chiroptical:
Very cool! Thanks @u-quark

quasi-coherent:
How is this meant to work?  I absolutely have a Makefile target called `gen`, but when I run this hook
```nix
gen = {
  enable = true;
  name = "Generate code from spec";
  entry = "make gen";
};
```
I get
```
Generate code from spec..........................................Failed
- hook id: gen
- exit code: 2

make: *** No rule to make target 'gen'.  Stop.
make: *** No rule to make target 'gen'.  Stop.
make: *** No rule to make target 'gen'.  Stop.
make: *** No rule to make target 'gen'.  Stop.
make: *** No rule to make target 'gen'.  Stop.
make: *** No rule to make target 'gen'.  Stop.
```

domenkozar:
You can specify directory with `make -C dir`

PR #97, opened 2021-04-05
Try fixing cargo check

### LINKED ISSUE #94: Cargo check fails to load crates.io-index correctly
When enabling `cargo-check.enable = true` I get this error:

```
cargo-check..............................................................Failed
- hook id: cargo-check
- exit code: 101

    Updating crates.io index
warning: spurious network error (2 tries remaining): [6] Couldn't resolve host name; class=Net (12)
warning: spurious network error (1 tries remaining): [6] Couldn't resolve host name; class=Net (12)
error: failed to get `base-x` as a dependency of package `hashexpr v0.1.0 (/build/src/hashexpr)`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  [6] Couldn't resolve host name; class=Net (12)

building '/nix/store/6lsmz02z9679x4wb1qiy0733689h2fk3-git-deps.drv'...
nix-linter...............................................................Passed
nixpkgs-fmt..............................................................Passed
shellcheck...........................................(no files to check)Skipped
builder for '/nix/store/wi0lbb6j045ma537m6xdvyrcya8g56rg-pre-commit-run.drv' failed with exit code 1
```
I guess the `class=Net` has gotten there by mistake?

Anderssorby:
Try to fix #94
nix-build does not allow for http access to ensure reproducibility. Since I were using the pre-commit hooks inside a CI which uses nix-build this will not work. So maybe we need something like https://github.com/nmattia/naersk to handle a pure package cache.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -200,7 +200,7 @@ in
         {
           name = "cargo-check";
           description = "Check the cargo package for errors";
-          entry = "${tools.cargo}/bin/cargo check";
+          entry = "${tools.cargo}/bin/cargo check --frozen";
           files = "\\.rs$";
           pass_filenames = false;
         };


domenkozar:
Does this solve your issue?

Anderssorby:
@domenkozar No. I'll look into this naersk solution when I get time.

domenkozar:
@Anderssorby why is `--frozen` needed then? Sorry, I'm not a Rust developer.

Anderssorby:
`--frozen` means no internet access which is necessary for nix-build, but requires that dependencies are already downloaded.

Anderssorby:
But most likely I will not want to merge this one because it doesn't solve the issue on it's own.

PR #101, opened 2021-04-26
Fixes #93

### LINKED ISSUE #93: Hpack does not seem to be able to deal with package.yaml files in subdirectories
```
hpack....................................................................Failed
- hook id: hpack
- exit code: 1

./package.yaml: Yaml file not found: /home/syd/src/sydtest/package.yaml
```

NorfairKing:
I tried it out on one of my projects with 
```
export NIV_OVERRIDE_nix_pre_commit_hooks=/home/syd/src/nix-pre-commit-hooks
```

CC @fiadliel

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -73,9 +73,8 @@ in
           name = "hpack";
           description =
             "hpack converts package definitions in the hpack format (package.yaml) to Cabal files.";
-          entry = "${tools.hpack}/bin/hpack --force .";
+          entry = "${tools.hpack-dir}/bin/hpack-dir";
           files = "(\\.l?hs$)|(^[^/]+\\.cabal$)|(^package\\.yaml$)";
-          pass_filenames = false;
         };
       ormolu =
         {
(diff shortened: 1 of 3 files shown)


NorfairKing:
I'm happy to take directions on the nix code here.
I used a new name: `hpack-dir` so that users can still add the `tools.hpack` to their shell and have it behave the way they think it will.

domenkozar:
I'm not sure this is the correct fix. It's going against how pre-commit works by listing files on its own instead of getting it from pre-commit.

What is triggering #93?

cc @fiadliel

NorfairKing:
@domenkozar I just followed the example of `terraform-fmt`. It also only lists files based on the arguments, which `pre-commit` passes in and will only be `package.yaml` files AFAICT.

domenkozar:
I see, I've missed the `$@` bit :) Looks good to me, hoping @fiadliel can take a look before I merge it.

domenkozar:
I've merged this, @fiadliel if you want to review it feel free to and we'll make the changes in a new PR.

PR #106, opened 2021-05-01
Use `types_or` instead of `types` for shellcheck

### LINKED ISSUE #105: shellcheck always  skipped
In `nix/default.nix` I have enabled shellcheck:
```nix
    pre-commit-check = (import sources."pre-commit-hooks.nix").run {
      inherit src;
      hooks = {
        shellcheck.enable = true;
        nixpkgs-fmt.enable = true;
        nix-linter.enable = true;
      };
```
Running `pre-commit run --all-files` just yields:
```
nix-linter...............................................................Passed
nixpkgs-fmt..............................................................Passed
shellcheck...........................................(no files to check)Skipped
```
although there is a bash script called `check-this` in my project.

Not sure where the problem is, but running `identify-cli check-this` yields:
```console
$ /nix/store/3h50fdl9wzr78gp9ykgs7d27a22a5yi4-python3.8-identify-1.5.5/bin/identify-cli  check-this 
["bash", "executable", "file", "shell", "text"]
```
So I think the script should be identified.

But changing the `types` settings from the list of "good shells" (https://github.com/cachix/pre-commit-hooks.nix/blob/f3fec488b5efed9104742811862e434e8f992dc9/modules/hooks.nix#L156-L165) to just
```
types: ["shell"]
```
(as found in the shellcheck hook here: https://github.com/jumanjihouse/pre-commit-hooks/blob/7cc5848088fd8412905ab79feea6c8edc3ac76c6/.pre-commit-hooks.yaml#L96)
makes it work:
```console
$ pre-commit run --all-files
nix-linter...............................................................Passed
nixpkgs-fmt..............................................................Passed
shellcheck...............................................................Failed
- hook id: shellcheck
- exit code: 1

In check-this line 5:
declare -r F=$( ls )
           ^-- SC2155: Declare and assign separately to avoid masking return values.


In check-this line 7:
echo $F
     ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
echo "$F"

For more information:
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

BTW, I also installed pre-commit via pip, latest version (2.12.1) thinking that it might identify files differently, but it also skipped the shellcheck hook.

Also created an issue here: https://github.com/pre-commit/pre-commit/issues/1896

avdv:
Fixes #105

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -153,6 +153,7 @@ in
         {
           name = "shellcheck";
           description = "Format shell files";
+          types = [ "shell" ];
           types_or =
             # based on `goodShells` in https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Parser.hs
             [
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -153,7 +153,7 @@ in
         {
           name = "shellcheck";
           description = "Format shell files";
-          types =
+          types_or =
             # based on `goodShells` in https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Parser.hs
             [
               "sh"
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -84,6 +84,15 @@ let
                   '';
                 default = [ "file" ];
               };
+            types_or =
+              mkOption {
+                type = types.listOf types.str;
+                description =
+                  ''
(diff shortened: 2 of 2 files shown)


avdv:
I have added the missing `types_or` option support. PTAL.

PR #111, opened 2021-06-04
Support per-hook stages

### LINKED ISSUE #110: Support specifying stages for individual hooks
Continuation of #49 

Currently, it's possible to specify `default_stages` for all hooks, which also configures which stages hooks are installed for.

Via the `raw` attribute on an individual hook, it's possible to specify `stages` for a single hook, though if any stage specified is not in the `default_stages`, the hook will not be installed.

So I suppose there are two aspects to this:

1. Add a `stages` config property on the hook level
2. Update `install_stages` to install both default stages (per current behaviour) and any specified in individual hooks

I'm happy to work on the implementation for this but it might be a few days before I get the time to do so, so I figured I'd post here first in case anyone has any comments.

sd234678:
Closes #110

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -14,7 +14,7 @@ let
   inherit (pkgs) runCommand writeText git;
 
   cfg = config;
-  install_stages = cfg.default_stages;
+  install_stages = lib.unique (cfg.default_stages ++ (builtins.concatLists (lib.mapAttrsToList (_: h: h.stages) enabledHooks)));
 
   hookType =
     types.submodule (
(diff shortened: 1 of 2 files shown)


roberth (changes requested):

sd234678:
Hey @roberth, I don't suppose you could re-review this?

modules/pre-commit.nix
```diff
@@ -117,12 +117,18 @@ let
                 description = "Whether to pass filenames as arguments to the entry point.";
                 default = true;
               };
+            stages =
+              mkOption {
+                type = types.listOf types.str;
+                description = "Confines the hook to run at a particular stage.";
+                default = cfg.default_stages;
```
roberth, line 124:
```suggestion
                defaultText = (lib.literalExpression or lib.literalExample) "default_stages";
```

roberth (commented):
Otherwise lgtm

modules/pre-commit.nix
sd234678: Thanks, I've updated that now to match

sd234678: Removing the value for `default` broke the build as stages was left without a value, but I've added the `defaultText` in addition. Was that what you were intending?

roberth: It is; I just used the suggestion feature wrong. Sorry about that.

sd234678: No worries, thanks for your help 😄

sd234678:
Hey @roberth, this all seems to be working okay now

roberth:
@domenkozar ^

PR #113, opened 2021-06-29
Ensure shellcheck runs even against files with no shebang

sd234678:
In some cases it makes sense to write shell files using shellcheck directives instead of shebangs, such as when the shebang is added by nix (like when using `writeShellScriptBin`).

This change allows those files to be picked up by shellcheck by default, since types_or and types are checked with an AND operator.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -179,6 +179,7 @@ in
               "bats"
               "dash"
               "ksh"
+              "shell"
             ];
           entry = "${tools.shellcheck}/bin/shellcheck";
         };


sd234678:
CI failures should be fixed by #114

domenkozar:
cc @roberth

roberth:
It seems that this would run shellcheck on unsupported script languages like zsh then.
Maybe it's best to just add a shebang. That clarifies the language (for other tools as well) and should work fine even if you're just prefixing a shebang. Only the first one counts. Any successive ones are just weird-looking `#` comments.

sd234678:
Thanks - I think you're right actually, this should probably be closed. I hadn't realised that successive ones were ignored even if they were in separate files.

roberth:
The [undocumented](https://github.com/NixOS/nixpkgs/issues/129741) `patchShebangs` bash function may also be an option. It's included in stdenv.

sd234678:
Ooh thanks, I'll look into that

PR #119, opened 2021-09-22
Removing local and passing silence option for hpack

maksar:
`local` keyword was producing an error on some bash
versions, being used out of the function scope,
decided to remove it.


domenkozar:
Is verbose the default?

PR #130, opened 2021-11-22
Flake module

roberth:
Done.

>   A module to import into flakes based on flake-parts.
>  Makes integration into a flake easy and tidy.
>  See https://github.com/hercules-ci/flake-parts#readme

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,4 +1,6 @@
+HOST_XDG_DATA_DIRS="${XDG_DATA_DIRS:-}"
 eval "$(lorri direnv)"
+export XDG_DATA_DIRS="${XDG_DATA_DIRS}:${HOST_XDG_DATA_DIRS}"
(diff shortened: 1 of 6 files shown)


domenkozar:
Hey Robert, do you plan to work on this one further?

roberth:
It's actually in a good state since the latest iteration. The remaining items weren't at all necessary and I've already incorporated feedback from Jonas.

roberth:
Rebased, ready and green :)

.envrc
```diff
@@ -1,4 +1,6 @@
+HOST_XDG_DATA_DIRS="${XDG_DATA_DIRS:-}"
```
domenkozar, line 1: I'd really like to get rid of lorri :( I think it's overengineered and with flakes not really needed.

PR #147, opened 2022-05-31
Add shfmt

loicreynier:
Adds `shfmt`, the shell formatter from [`mvdan/sh`](https://github.com/mvdan/sh).

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ eval "$shellHook"
 
 - [prettier](https://prettier.io)
 - [clang-format](https://clang.llvm.org/docs/ClangFormat.html)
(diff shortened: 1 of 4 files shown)


loicreynier:
You are right, adding  `types_or` removes `.sh` files detection. I had copied this from the `shellcheck` hook which should have the same issue. I removed it and it works like a charm. Other hooks do not use it either, see [this one](https://github.com/jumanjihouse/pre-commit-hooks/blob/9daf4ccb1f73e83c75517076bc2c1aafde20a082/.pre-commit-hooks.yaml#L117) as an example.

PR #151, opened 2022-06-15
Preserve executable bit in files

tljuniper:
This changes the pre-commit-run to preserve the executable bit of the files-to-check.

Some commit hooks such as `pre-commit-hooks/check-executables-have-shebangs` are designed to run only on executable files via `types = ["executable"]` and the python tool identify uses the executable bit of a file when generating file tags, so `executable` is only listed when that bit is set.

If we don't preserve the executable bit, results of a local pre-commit invocation and the `pre-commit-check` might differ.

This was easy to test using the `check-executables-have-shebangs` hook:

```nix
      check-executables-have-shebangs = {
        enable = true;
        name = "Check that executables have shebangs";
        description = "Ensures that (non-binary) executables have a shebang.";
        entry = "${pkgs.python3Packages.pre-commit-hooks}/bin/check-executables-have-shebangs";
        types = [ "text" "executable" ];
      }
```

The git output of `run` is very helpful as well because it shows the file mode.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -163,7 +163,10 @@ let
     runCommand "pre-commit-run" { buildInputs = [ git ]; } ''
       set +e
       HOME=$PWD
-      cp --no-preserve=mode -R ${cfg.rootSrc} src
+      # Use `chmod +w` instead of `cp --no-preserve=mode` to be able to write and to
+      # preserve the executable bit at the same time
+      cp -R ${cfg.rootSrc} src
+      chmod -R +w src
       ln -fs ${configFile} src/.pre-commit-config.yaml
       cd src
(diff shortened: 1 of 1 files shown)


m1-s:
@domenkozar Can this be merged?

gkze:
I think this is happening again. I have a Python file with `#!/usr/bin/env python` that is executable in the worktree, and it passes `ruff check`, but when I run `nix flake check` it complains with [`EXE001`](https://docs.astral.sh/ruff/rules/shebang-not-executable/). Looks like things have changed since this was merged.

I'd love to contribute a fix, but not sure where to look, as I'm not too familiar with how the source tree is copied into the store - if anyone can point me in the right direction, it would be appreciated 🙏🏻

PR #152, opened 2022-06-15
Add markdownlint

tljuniper:
Add support for markdownlint.

The markdownlint auto-formatter also fixed some small things in the README (the rationale for removing the leading `$ ` in shell snippets without any command output is that they are easier to copy this way).


domenkozar:
Could you rebase on top of master?

tljuniper:
Rebase is done.

PR #165, opened 2022-09-26
feat: add actionlint, luacheck, chktex, latexindent

lovesegfault:
I also did a drive-by sort of the input attrset of `tools.nix` as it was getting
really messy


lovesegfault:
Tagging @domenkozar for review + letting CI run

PR #177, opened 2022-10-21
feat: add govet hook

roman:
Hello :wave:

I want to introduce various golang tools to the `pre-commit-hooks` project. This is my first stab at the challenge.

### Questions

I've a few questions regarding the contribution guidelines:

>   Nix closure of the tool should be small e.g. < 50MB

What command can I use to validate that I'm not violating this constraint?

> The tool needs to live in a separate repository (even if a simple bash script, unless it's a oneliner) 

I think I'm cheating here because I include an inline bash code in the config of the hook; if this is not acceptable, can you provide an example of how we include a separate repository? Should I create a repo with bash scripts and have it as a flake input? or would you prefer a different approach?

Would you mind if I attempt to write a setup for running NixOS Tests that exercises pre-commit hooks? Or would you rather not have that?

Thanks.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -494,5 +494,26 @@ in
           entry = "${tools.markdownlint-cli}/bin/markdownlint";
           files = "\\.md$";
         };
+
+      govet =
+        let
+          script = pkgs.writeShellScript "precommit-govet" ''
+            for dir in $(echo "$@" | xargs -n1 dirname | sort -u); do
+                ${tools.go}/bin/go vet ./"$dir"
+            done
+          '';
+        in
+        {
+          name = "govet";
+          description = "Checks correctness of Go programs";
+          entry = builtins.toString script;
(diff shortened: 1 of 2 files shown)


domenkozar:
Added https://github.com/cachix/pre-commit-hooks.nix/commit/364f8c3c66f4326992264fac489a0d4e634248d6 to elaborate on that a bit. I wonder how comes go is so big?

> Would you mind if I attempt to write a setup for running NixOS Tests that exercises pre-commit hooks? Or would you rather not have that?

Do we need nixos tests? I was hoping we can run those test just part of the github actions.

domenkozar:
Actually given that we now have lazy attribute set (meaning that only tools that are used will get fetched) it's up to the user to decide what hooks are good enough :)

domenkozar:
Could you also open a PR to add it to the README? Thanks!

PR #183, opened 2022-11-05
add treefmt

terlar:
This is useful to abstract away all the language specific formatting hooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ use nix
 - [editorconfig-checker](https://github.com/editorconfig-checker/editorconfig-checker)
 - [actionlint](https://github.com/rhysd/actionlint)
 - [tagref](https://github.com/stepchowfun/tagref)
+- [treefmt](https://github.com/numtide/treefmt)
 
 You must configure which languages should be formatted by `clang_format` using
 `clang-format.types_or`. For example to check both C and C++ files:
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -13,7 +13,7 @@ let
 in
 {
   options = {
-    perSystem = flake-parts-lib.mkPerSystemOption ({ config, pkgs, ... }:
+    perSystem = flake-parts-lib.mkPerSystemOption ({ config, options, pkgs, ... }:
       let cfg = config.pre-commit;
       in
       {
@@ -63,6 +63,7 @@ in
             rootSrc = self.outPath;
             package = lib.mkDefault pkgs.pre-commit;
             tools = import ./nix/call-tools.nix pkgs;
+            settings.treefmt.package = lib.mkIf (options?treefmt) (lib.mkDefault config.treefmt.build.wrapper);
           };
         };
       });
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -368,6 +368,34 @@ in
           };
         };
 
+      treefmt =
+        {
+          package = mkOption {
+            type = types.package;
+            description = lib.mdDoc
+              ''
+                The `treefmt` package to use.
+
+                Should include all the formatters configured by treefmt.
+
+                For example:
+                ```nix
+                pkgs.writeShellApplication {
+                  name = "treefmt";
+                  runtimeInputs = [
+                    pkgs.treefmt
+                    pkgs.nixpkgs-fmt
+                    pkgs.black
+                  ];
+                  text =
+                    '''
+                      exec treefmt "$@"
+                    ''';
+                }
+                ```
+              '';
+          };
+        };
     };
 
   config.hooks =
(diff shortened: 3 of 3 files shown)


roberth:
How does treefmt get its tools?
As is, this hook seems to beincompatible out of the box with `checks.<system>.pre-commit`.
Assuming the deps come from a devShell, perhaps we could add a convenience option to the flakeModule to automate getting deps from a devShell.

terlar:
Hmm, good point! I thought this was working, seems perhaps the `treefmt` command is silently passing in this case. I tried configure this and the `checks.<system>.pre-commit` passes even when it shouldn't.

It would need to get all the tools configured. I am not sure if everyone will get the deps from a devShell, but that is what I usually do.

Will have to think a bit more about this, not as straight-forward as I thought...

terlar:
One alternative would be to pass a tool with wrapped dependencies also. Basically a treefmt with all the tools on `PATH` (from Nix store of course).

terlar:
Also to make it even more fancy the `treefmt.toml` could be parsed and only the configured files would be included, to avoid having it run everytime, only when file-types configured has changed.

terlar:
So I have verified that a wrapped `treefmt` is working well. I see three stages of `treefmt` support:

- Stage 1: Letting the user provide a wrapper via tools (current state of this PR)
- Stage 2: Moving the wrapper into this project and letting the user specify which formatters their treefmt needs
- Stage 3: Parsing the `treefmt.toml` to automatically infer which formatters are used, I guess the user would need something similar to tools or if we re-use tools to define where the packages originate from.

What do you think, is it worth to merge this as a start. Do we want to support these various stages?

Here is a currently working usage example (using flake-parts with this):
```nix
{
  perSystem = { pkgs, ... }: {
    pre-commit = {
      check.enable = true;
      pkgs = pkgs // {
        treefmt = pkgs.writeShellApplication {
          name = "treefmt";
          runtimeInputs = [
            pkgs.treefmt
            # Formatters
            pkgs.alejandra
            pkgs.black
          ];
          text = ''
            treefmt "$@"
          '';
        };
      };
      settings.hooks = {
        treefmt.enable = true;
      };
    };
  };
}
...
```

terlar:
Hmm, or seems pkgs override via flake-parts is not working. Will try to create a minimal reproducible flake and open a ticket for that

terlar:
I am currently using this kind of configuration together with [treefmt-nix](https://github.com/numtide/treefmt-nix) which also has flake-parts integration:
```nix
{
  perSystem = { pkgs, config, ... }: {
    treefmt.programs.alejandra.enable = true;

    pre-commit = {
      check.enable = true;
      pkgs = pkgs.extend (final: prev: {
        treefmt = config.treefmt.build.wrapper;
      });
      settings.hooks = {
        treefmt.enable = true;
      };
    };
  };
}
```

I think this interface is fine enough. It supports both treefmt-nix and custom wrapper as mentioned before. Since the interface of package customization is like that currently I don't see a need to further complicate things. If we don't want to pull in `treefmt-nix` as a submodule for this integration. E.g. `pre-commit.settings.settings.treefmt.programs.alejandra.enable = true`.

roberth:
If you add a `settings.treefmt.package` option, you don't need the overlay, and you could set a `lib.mkDefault` in `flake-module.nix`.

You can detect whether the module for `treefmt-nix` is loaded with the `options` parameter. Something like `pre-commit.settings.settings.treefmt.package = mkIf (options?treefmt) config.treefmt.build.wrapper`. It might even be ok to check `config` in this case, but picking the less dynamic `options` helps with avoiding infinite recursion in some (other) cases.

terlar:
@roberth Thank you, this worked out really well. I think this is quite nice behavior. So now it will fail by default if you try to enable treefmt with explicitly setting the package (if you don't use the flake-parts treefmt-nix, then it will add it automatically.

srid:
I confirm this to work (with `treefmt-nix` flake-parts module) via:

```nix
        pre-commit = {
          check.enable = true;
          settings.hooks = {
            treefmt.enable = true;
          };
        }
```

@domenkozar Could we get this merged please? Seems like a fairly simple change.

roberth (approved):

jh-devv:
@terlar Hey! Sorry to bump this thread, but how do I know that `treefmt` is *doing it's thing*
image
I mean, should it show the output of the linters enabled by `treefmt`?

terlar:
Yes as the author of `pre-commit` indicates it is silent by default unless the pre-commit command returns a non-zero exit code.

You can see the full output by:
```
pre-commit run --verbose
```
or if you want to force everything:
```
pre-commit run --all-files --verbose
```

I think this is because the cache invalidation error is not seen as a critical error worthy of returning non-zero exit code in treefmt.

If treefmt shows Passed it means it did it's thing.

Perhaps if cache is causing problems we could add `--no-cache`. Perhaps we should always do that since `pre-commit` has it's own kind of cache (e.g. only checking files that changed).

PR #191, opened 2022-11-20
Use `-d` to check for `.git` instead of `-f`

cpcloud:
This PR fixes a bug where the config isn't being installed due to an
erroneous check of `.git` using `-f` instead of `-d`.


roberth:
Should be probably be `-e` (exists), because worktrees make use of a `.git` file instead of a directory.

PR #202, opened 2022-11-30
Fix Prettier output option

loicreynier:
Fix Prettier output option that I did not properly tested before creating #201, sorry about that.


loicreynier:
#203 fixes the same thing in addition to a spelling mistake in `typos` hook entry.

roberth (approved):

PR #203, opened 2022-12-01
Fix typo

roberth:
Fixes docs generation.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -145,7 +145,7 @@ in
           output =
             mkOption {
               description = lib.mdDoc "Output format.";
-              type = types.nullOr types.enum [ "check" "list-diffrent" ];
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -145,7 +145,7 @@ in
           output =
             mkOption {
               description = lib.mdDoc "Output format.";
-              types = types.nullOr types.enum [ "check" "list-diffrent" ];
+              type = types.nullOr types.enum [ "check" "list-diffrent" ];
```
loicreynier, line 148:
```suggestion
              type = types.nullOr types.enum [ "check" "list-different" ];
```

There is another typo on this line in the `enum`, my bad.

roberth:
Didn't look for #202 when fixing this trivial mistake. Closing in favor of that.

PR #206, opened 2022-12-06
Add Commitizen

loicreynier:
Add [Commitizen](https://github.com/commitizen-tools/commitizen) Git commit message check.

Closure size is 580K according to `du -sh $(nix-build -A commitizen)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,10 @@ clang-format = {
 };
 ```
 
+## Git
+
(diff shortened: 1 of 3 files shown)


domenkozar:
Seems like it doesn't build on darwin because tests fail.

loicreynier:
Rebased into master and disabled Commitizen build tests.

PR #207, opened 2022-12-06
Add Pylint & flake8

loicreynier:
Add [Pylint](https://github.com/PyCQA/pylint) with options to control output format.


loicreynier:
Just added flake8 in the same PR since both are related.

PR #210, opened 2022-12-12
Fix clang-format types

loicreynier:
Fix `clang-format` hooks type so it doesn't try to format not supported files.


roberth:
Feel free to add the link in an inline comment. That will help with changes and reviews later.

loicreynier:
Good point, permalink added.

PR #211, opened 2022-12-13
Haskell formatters: Include .hs-boot files

shlevy:


shlevy:
Note: I've only tested this with `fourmolu`

shlevy:
Also with `hlint`

Issue #208, opened 2022-12-07
Add pre-commit-hook-ensure-sops?


nialov:
Hey,

I have packaged and configured [`pre-commit-hook-ensure-sops`](https://github.com/yuvipanda/pre-commit-hook-ensure-sops) that checks the repository for improperly managed `sops` secrets. It is useful for paranoid [`sops-nix`](https://github.com/Mic92/sops-nix) users.

See here: https://gist.github.com/nialov/867d96761f2a68168e09a10d8fe6f63a

Should I do a pull request or are these smaller, quite specialized hooks, wanted in `pre-commit-hooks.nix`?

domenkozar:
Would be good to have it here!

nialov:
Cool, I will make a pull request (hopefully) over the weekend.

PR #212, opened 2022-12-16
Add pre-commit-hook-ensure-sops

nialov:
As suggested (by me) here #208, I have implemented [`pre-commit-hook-ensure-sops`](https://github.com/yuvipanda/pre-commit-hook-ensure-sops) as a hook in this pull request.

Some notes/questions/problems:

- There is no separate repository where the package itself is in (e.g., `nixpkgs`) so I have packaged it in `./nix/tools/pre-commit-hook-ensure-sops`. I suppose this is not ideal (?). If it first should be added in `nixpkgs` I can do the pull request there.
- The name is pretty verbose but matches the name of the original repository. Could be changed but then there is the chance for confusion of what the hook is based on originally.
- The secret files made by `sops` can have (any) file extensions and part of the check is making sure all files are valid `yaml` (i.e., `json` is accepted). Therefore, I do not know of a way of how to target the hook by default. Currently I implemented it to target all files within a `secrets` directory with the pattern `^secrets`. This unlikely matches the namings of other repositories but then the users can just override the value. Any other options or suggestions?

Let me know what you think! I will modify the pull request as suggested.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -56,11 +56,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1685866647,
-        "narHash": "sha256-4jKguNHY/edLYImB+uL8jKPL/vpfOvMmSlLAGfxSrnY=",
+        "lastModified": 1689261696,
+        "narHash": "sha256-LzfUtFs9MQRvIoQ3MfgSuipBVMXslMPH/vZ+nM40LkA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "a53a3bec10deef6e1cc1caba5bc60f53b959b1e8",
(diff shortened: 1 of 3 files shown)


domenkozar:
Let's add it to nixpkgs and use it from there please :)

nialov:
Roger! Will submit a pull request in `nixpkgs` at some point. Probably start of next year at the earliest.

nialov:
The package is now in `nixpkgs` but the project repo flake `nixpkgs` input is not yet updated. Therefore the pull request contains an update to `nixpkgs` input in `flake.lock` which is probably undesirable.

domenkozar:
You'll also need to `nix flake update`

domenkozar:
There are a few hooks that conditionally work on 23.05, you'll need to handle this one similarly.

PR #213, opened 2022-12-18
add php-cs-fixer as hook

shyim:
Adds PHP support :)

image


shyim:
font forge seems broken on macos https://github.com/cachix/pre-commit-hooks.nix/actions/runs/3725555635/jobs/6318625636#step:5:2239

PR #221, opened 2023-01-05
Add option to fail when clippy emits warnings

notgne2:
This makes it possible for projects to use clippy to enforce lints without using `#![deny(warnings)]` (considered harmful ™️)


domenkozar:
Could you resolve conflicts?

PR #225, opened 2023-01-11
fix: allow .git directory on parent directory

### LINKED ISSUE #224: Support for mono-repos?
### Context

As a developer that works on a project inside a monorepo
I would like to inject a pre-commit hook on a subdirectory of the git repo
So that I can take advantage of the fantastic pre-commit-hooks.nix tool

### Problem

When we have a `flake.nix` file with a pre-commit setup that runs the `shellHook` on a sub-directory of the repository, the `pre-commit-hooks.nix` tool doesn't install the pre-commit hooks. Instead, it reports:

> WARNING: pre-commit-hooks.nix: .git not found; skipping installation.

### Current behavior

It seems `pre-commit-hooks.nix` assumes that the `flake.nix` file and the `.git` folder are in the same directory.

### Expected behaviour

Ideally, we can get a pre-commit-hook installation that works only on the sub-directory of the repository that contains the flake. 

Is this possible?

roman:
Currently, the check to validate pre-commit-hooks is getting installed on a git repository only considers the `.git` directory if the pre-commit-hooks config is at the root path; if the pre-commit-hooks config lives in another path that is not the root of the repository, pre-commit-hooks won't install a script.

This change allows `flake.nix` files in sub-directories to have the pre-commit-hooks configuration. This is especially common in monorepos.

Note, this change requires a [PR on `nixpkgs`](https://github.com/NixOS/nixpkgs/pull/210256) to be merged before this code works.

resolves #224

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -366,7 +366,7 @@ in
           if ! type -t git >/dev/null; then
             # This happens in pure shells, including lorri
             echo 1>&2 "WARNING: pre-commit-hooks.nix: git command not found; skipping installation."
-          elif [[ ! -e .git ]]; then
+          elif ! ${git}/bin/git rev-parse --git-dir &> /dev/null; then
             echo 1>&2 "WARNING: pre-commit-hooks.nix: .git not found; skipping installation."
           else
             # These update procedures compare before they write, to avoid


roman:
Hey @domenkozar, this has been here for a while, and recently started working due to the dependency being merged to nixpkgs, would you consider adding it?

domenkozar:
@roman could you backport it to 22.11? We also want to support the stable version in pre-commit-hooks.nix

modules/pre-commit.nix
```diff
@@ -366,7 +366,7 @@ in
           if ! type -t git >/dev/null; then
             # This happens in pure shells, including lorri
             echo 1>&2 "WARNING: pre-commit-hooks.nix: git command not found; skipping installation."
-          elif [[ ! -e .git ]]; then
+          elif ! ${pkgs.locate-dominating-file}/bin/locate-dominating-file .git >/dev/null; then
```
imsofi, line 369: This should probably be written with `git rev-parse --git-dir` instead of using `locate-dominating-file` which may fail or become slow at deep folder structures due to the expectation to find a folder named `.git`.

roman: I was trying to understand the nix runtime dependencies of this generated script, but I'm unable to find any via `nix-store --query`. 

We must have the assumption that `git` is installed already, so I don't think avoiding depending on it gains us anything in this context.

All this to say, I agree @imsofi, let me change that to the `git` command.

PR #226, opened 2023-01-12
feat: add bats checker

roman:
### Context

As a bash developer
I want to run tests on pre-commit
So that I don't introduce bugs in my bash logic

### Summary

This extension allows us to run the [`bats`](https://bats-core.readthedocs.io/en/stable/) utility.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -605,6 +605,14 @@ in
             ];
           entry = "${tools.shellcheck}/bin/shellcheck";
         };
+      bats =
+        {
+          name = "bats";
+          description = "Run bash unit tests.";
+          types = [ "shell" ];
+          types_or = [ "bats" "bash" ];
+          entry = "${tools.bats}/bin/bats -p";
+        };
       stylua =
         {
           name = "stylua";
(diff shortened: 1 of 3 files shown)


modules/hooks.nix
```diff
@@ -605,6 +605,17 @@ in
...
           entry = "${tools.shellcheck}/bin/shellcheck";
         };
+      bats =
+        {
+          name = "bats";
+          description = "Run bash unit tests.";
+          types = [ "shell" ];
+          types_or =
+            [
+              "bats"
```
loicreynier, line 615:
```suggestion
              "bash"
```

Shouldn't it be `bash` instead of `bats` in `types_or`?

loicreynier (changes requested):

modules/hooks.nix
roman: I'm not sure; I understand that this list the files that trigger this pre-commit hook.  Is this the case?

If that is the case, I wouldn't like to get all `.bash` files to get called with bats because it is very likely those are not bats test. By having only `bats` here, the changes to `.bats` file are the only ones that will get picked up by the hook.

roman: On second thought, I tested bats running on regular bash files and it ignores them correctly, so, sure let me add that

roman:
@loicreynier change is done

roman:
Is anything else needed to move this along @domenkozar?

PR #227, opened 2023-01-12
Add Ruff

shikanime:
Ruff is a Python linter that is starting to get a lot of attention lately.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -353,7 +353,7 @@ in
         {
           name = "ruff";
           description = " An extremely fast Python linter, written in Rust.";
-          entry = "${pkgs.python3Packages.black}/bin/ruff";
+          entry = "${pkgs.ruff}/bin/ruff";
           types = [ "python" ];
         };
       cabal2nix =
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -354,7 +354,7 @@ in
           name = "ruff";
           description = " An extremely fast Python linter, written in Rust.";
           entry = "${pkgs.python3Packages.black}/bin/ruff";
-          types = [ "file" "python" ];
+          types = [ "python" ];
         };
       cabal2nix =
         {
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -349,6 +349,13 @@ in
           entry = "${pkgs.python3Packages.black}/bin/black";
           types = [ "file" "python" ];
         };
+      ruff =
+        {
+          name = "ruff";
+          description = " An extremely fast Python linter, written in Rust.";
+          entry = "${pkgs.python3Packages.black}/bin/ruff";
+          types = [ "file" "python" ];
+        };
       cabal2nix =
         {
           name = "cabal2nix";
--- a/nix/tools.nix
+++ b/nix/tools.nix
@@ -29,6 +29,7 @@
 , nodePackages
 , ormolu
 , python39Packages
+, ruff
 , runCommand
 , rustfmt
 , shellcheck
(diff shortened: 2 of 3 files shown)


modules/hooks.nix
```diff
@@ -359,6 +359,13 @@ in
           entry = "${pkgs.python3Packages.black}/bin/black";
           types = [ "file" "python" ];
         };
+      ruff =
+        {
+          name = "ruff";
+          description = " An extremely fast Python linter, written in Rust.";
+          entry = "${pkgs.python3Packages.black}/bin/ruff";
+          types = [ "file" "python" ];
```
loicreynier, line 367:
```suggestion
          types = [ "python" ];
```

The type `file` is not necessary when using the `python` type.

modules/hooks.nix
```diff
@@ -359,6 +359,13 @@ in
           entry = "${pkgs.python3Packages.black}/bin/black";
           types = [ "file" "python" ];
         };
+      ruff =
+        {
+          name = "ruff";
+          description = " An extremely fast Python linter, written in Rust.";
+          entry = "${pkgs.python3Packages.black}/bin/ruff";
```
loicreynier, line 366:
```suggestion
          entry = "${pkgs.ruff}/bin/ruff";
```

`ruff` is not a Python package.

loicreynier (changes requested):

nix/tools.nix
```diff
@@ -53,7 +53,7 @@ let
   };
 in
 {
-  inherit actionlint alejandra cabal-fmt cabal2nix cargo clang-tools clippy deadnix dhall editorconfig-checker hadolint hindent hlint hpack html-tidy nix-linter nixfmt nixpkgs-fmt ormolu rustfmt shellcheck shfmt statix stylish-haskell stylua typos go mdsh revive go-tools;
+  inherit actionlint alejandra cabal-fmt cabal2nix cargo clang-tools clippy deadnix dhall editorconfig-checker hadolint hindent hlint hpack html-tidy nix-linter nixfmt nixpkgs-fmt ormolu rustfmt shellcheck shfmt statix stylish-haskell stylua typos go mdsh revive go-tools ruff;
```
loicreynier, line 56: `ruff` should also be added to the arguments line 32.

loicreynier (changes requested):

shikanime:
Oh, at the moment, it seems that Ruff is only available on Nixpkgs version 22.11.

domenkozar:
I'll try to bump checks to 22.11

domenkozar:
If you rebase it should work now

dhess:
FYI, I believe this commit broke the flake. We now get:

```
error: Function called without required argument "ruff" at /nix/store/lvwknwz9wfx66qyxdcak5p6w2wfxlzpd-source/nix/tools.nix:32, did you mean "duff", "guff" or "rufo"?
```

shikanime:
Please correct me if I'm wrong, as the package is only available in 22.11 release, should I have added a default to null ?

dhess:
@shikanime Possibly. This happens when I use `pre-commit-hooks.nix` with `haskell.nix`, which is, I believe, using a pre-22.11 nixpkgs-unstable pin. There's a PR to bump it: https://github.com/input-output-hk/haskell.nix/pull/1815

shikanime:
Thanks for the extra details, I didn't realize the follows mechanism worked like that after reading the documentation.
I haven't had much experience with Flake dependencies system.
Sorry for the inconveniences (⁠｡⁠ŏ⁠﹏⁠ŏ⁠)

dhess:
No worries at all, thanks for the contribution.

PR #229, opened 2023-01-17
feat: add autoflake

yajo:
Very useful in Python projects.

@moduon MT-1075

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -306,6 +306,26 @@ in
             };
         };
 
+      autoflake =
+        {
+          binPath =
+            mkOption {
+              type = types.str;
+              description = lib.mdDoc "Path to autoflake binary.";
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -952,6 +972,14 @@ in
           types = [ "python" ];
         };
 
+      autoflake = 
+        {
+          name = "autoflake";
+          description = "Remove unused imports and variables from Python code.";
+          entry = "${settings.autoflake.binPath} ${settings.autoflake.flags}";
+            types = [ "python" ];
```
domenkozar, line 980:
```suggestion
          types = [ "python" ];
```

domenkozar:
seems like there's an extra whitespace

yajo:
It should be OK now.

PR #233, opened 2023-01-25
fix(pre-commit-run): Remove .git directory

nicknovitski:
It seems like this is the intention of the code, but currently the `rm` command deletes nothing, and doesn't fail because it is given the `-f` flag.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -174,7 +174,7 @@ let
       chmod -R +w src
       ln -fs ${configFile} src/.pre-commit-config.yaml
       cd src
-      rm -rf src/.git
+      rm -rf .git
       git init
       git add .
       git config --global user.email "you@example.com"


roberth:
Actually the existence of a `.git` indicates that you're probably going to have excessive rebuilds and store garbage. The source attribute should be pre-filtered to exclude `.git` and the .gitignored files. Maybe it should be an error to even have a `.git` in the source?

domenkozar:
The code does `cp -R ${cfg.rootSrc} src` so there should/could be a `.git` in src. Can you show us how you're using pre-commit-run?

nicknovitski:
> The code does `cp -R ${cfg.rootSrc} src` so there should/could be a `.git` in src.

There is a `.git` in `src`, but the line before `rm` is `cd src`, see?  So if you want to delete `src/.git`, after running `cd src`, you have to run `rm -r .git`, not `rm -r src/.git`.

nicknovitski:
Let me try to make sure I've explained this.

Currently the lines are:
```
cd src
rm -rf src/.git
```

My proposed change is:
```
cd src
rm -rf .git
```

Or equivalently it could be:
```
rm -rf src/.git
cd src
```

Or, if you prefer to keep the directory:
```
cd src
```

lafrenierejm (approved):

modules/pre-commit.nix
```diff
@@ -174,7 +174,7 @@ let
       chmod -R +w src
       ln -fs ${configFile} src/.pre-commit-config.yaml
       cd src
-      rm -rf src/.git
+      rm -rf .git
```
roberth, line 177:
```suggestion
      if [[ -e .git ]]; then
        echo 'error: The src of your pre-commit check contains a .git directory. Please filter it out, so that the check is more reproducible and more efficient with store resources.'
        exit 1
      fi
```

roberth (changes requested):
I think having a .git in the sandbox is always bad, and we should help users prevent that.
Admittedly the message could be more helpful about how to filter it out, but this suggestion would be a start.

PR #242, opened 2023-02-12
Add luacheck, chktex, and latexindent to README

lunik1:
Some checkers introduced in #165 were not added to the README. This patch adds them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,7 @@ use nix
 
 ## Lua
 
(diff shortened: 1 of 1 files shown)


lunik1:
Just noticed that the `--modifyIfDifferent` flag does not work for `latexindent`. It should be `--overwriteIfDifferent`. Should I fix that here or in a new PR?

domenkozar:
Best to do a new PR :)

PR #248, opened 2023-02-23
Add mypy support

mweinelt:
The mypy tool provides static type checks for python and is very common in modern python projects.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -118,6 +118,18 @@ in
             default = { };
           };
       };
+      mypy =
+        {
+          binPath =
+            mkOption {
+              type = types.str;
+              description = lib.mdDoc "Mypy binary path. Should be used to specify the mypy executable in an environment containing your typing stubs.";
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -573,6 +573,13 @@ in
           entry = toString script;
           files = "\\.md$";
         };
+      mypy =
+        {
+          name = "mypy";
+          description = "Static type checker for Python";
+          entry = "${pkgs.mypy}/bin/mypy";
```
loicreynier, line 580: We should add an option to set a custom binary path. `mypy` needs to be ran from an environment where all the Python packages with custom types are installed, i.e. your dev environment. See the `pylint.binPath` option for reference.

loicreynier (changes requested):

modules/hooks.nix
mweinelt: Thanks, added a `binPath` option as advised.

PR #254, opened 2023-03-15
Add a hook for Topiary

Niols:
As of yesterday, Topiary 0.1.0 has made it into `nixpkgs` (cf https://github.com/NixOS/nixpkgs/pull/220980). This pull request adds support for a Topiary pre-commit hook.

The CLI has changed at some point between the previous version in `nixpkgs` and 0.1.0, so the intermediary Shell script `topiary-inplace` will not work with an older version of `nixpkgs`. I don't really know what to do with this: if we merge this PR today, the Topiary pre-commit hook will be unusable out of the box (and will only work if someone puts eg. `inputs.pre-commit-hooks.inputs.nixpkgs.follows` to a more recent version of `nixpkgs`). Should we just wait for Topiary to be a bit more established in `nixpkgs`? Should I also `nix flake lock --update-input nixpkgs` in this PR? What would be the right thing to do?


Niols:
I guess it's without surprise that the tests don't pass with error:
```
Function called without required argument "topiary" 
```

Niols:
Rebased on `master` to fix the conflict in `README.md`.

domenkozar:
You can see how some other tools allow null package so that they work even if the tool is not in stable.

PR #257, opened 2023-03-16
Markdownlint fix by default

shikanime:
For the sake of convenience, it would be better if markdownlint fixed the problems that can be fixed such as Eslint and Gofmt.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -732,7 +732,7 @@ in
         {
           name = "markdownlint";
           description = "Style checker and linter for markdown files.";
-          entry = "${tools.markdownlint-cli}/bin/markdownlint -c ${pkgs.writeText "markdownlint.json" (builtins.toJSON settings.markdownlint.config)}";
+          entry = "${tools.markdownlint-cli}/bin/markdownlint --fix -c ${pkgs.writeText "markdownlint.json" (builtins.toJSON settings.markdownlint.config)}";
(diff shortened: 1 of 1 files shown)


aaronmondal:
@shikanime Sorry for the late reply I didn't see this 😳

It seems like this specific case was an issue for a long time https://github.com/DavidAnson/markdownlint/issues/207. I haven't tried the solutions proposed in that issue, but if I understood correctly there should be *some* way to make things work with custom syntax extensions. I wonder whether one of the most used doc generators should really be treated as "custom" though 😅.

aaronmondal:
I should note that the current behavior can be worked around by doing something like this:

https://github.com/eomii/rules_ll/blob/8102c7d81093469dbd53e8d3ddb49b6702f52d13/docs/guides/cuda_and_hip.md?plain=1#L10-L14

Even with this bug its prooobably ok to make this change. I'd imagine that those using linters/formatters a lot would care enough about style issues that they'd notice formatting issues early and just work around it.

domenkozar:
What's the conclusion here?

totoroot:
I could think of a bunch of occasions where fixing by default would cause issues, so I'll vote for not changing the behaviour.
Like @shikanime has said, after all it's a linter.

shikanime:
On reflection, the community seems to be resisting this change with understandable arguments.
I think closing this RP is a reasonable conclusion at this time.

PR #261, opened 2023-03-17
Make hook installation optional

roberth:
A highly opinionated workflow doesn't work for all projects, especially ones that already have a long history, notably NixOS/nix. Some people hate pre-commit with a passion.

These things should not be showstoppers.

To quote the docs:

## Opting out, opting in

Users can enforce their own preference by setting `NIX_PRE_COMMIT_HOOKS_INSTALL` to `0` or `1` in their environment, e.g. in `.profile` or in [`home.sessionVariables`](https://nix-community.github.io/home-manager/options.html#opt-home.sessionVariables) for home-manager users.

Furthermore, project maintainers can set a default preference via the `installByDefault` option.

Finally, regardless of these settings, you may install the hooks manually by running `nix-pre-commit-hooks-install` in your shell. This does what it says, and the shell hook will detect that the hooks exist and keep them up to date.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -379,67 +379,9 @@ in
           _pre_commit_hooks_nix_config=${configFile}
           _pre_commit_hooks_nix_install_stages='${concatStringsSep " " install_stages}'
 
-          if ! type -t git >/dev/null; then
-            # This happens in pure shells, including lorri
-            echo 1>&2 "WARNING: pre-commit-hooks.nix: git command not found; skipping installation."
-          elif ! $_pre_commit_hooks_nix_git rev-parse --git-dir &> /dev/null; then
-            echo 1>&2 "WARNING: pre-commit-hooks.nix: .git not found; skipping installation."
-          else
-            GIT_WC=`$_pre_commit_hooks_nix_git rev-parse --show-toplevel`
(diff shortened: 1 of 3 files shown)


domenkozar:
I'm confused why this is needed, can't you make the call to pre-commit `run` optional based on a boolean?

roberth:
I don't think `run` is responsible for installation.

The goal is to let a project choose not to install any hooks into .git by default, unless the devShell user has opted in.
I think *something* might done at the project level, but it won't be quite as user friendly, as it doesn't standardize this installation logic.

roberth:
I've re-done it. This time it just (cleanly) refactors the existing logic, so that it can be called without installing the hooks into `.git`.
Solved some tech debt along with that. We now have shellcheck on the script.

phip1611:
@roberth I'd love to have this. Are you willing to refresh your PR?

domenkozar:

domenkozar:
Closing as outdated

phip1611:
Is the functionality request outdated or just the state of the PR?

Is there a replacement?

domenkozar:
I think having `enable` option would make this a lot simpler, I'm happy to merge it if someone makes a PR.

domenkozar:
https://github.com/cachix/git-hooks.nix/pull/610

PR #266, opened 2023-03-30
flake-parts: add `devShell` option

srid:
Having flake-parts modules expose a `devShell` like this is really useful when the user has a top-level devShell and they want to "compose" other devShells using `inputsFrom`.

cf. https://github.com/srid/haskell-flake/issues/135

cc @roberth


srid:
@domenkozar Looks like the CI failures are unrelated.

PR #268, opened 2023-03-31
Recognize `$FLAKE_ROOT` as location of `.pre-commit-config.yaml`

### LINKED ISSUE #267: `.pre-commit-config.yaml` is not always installed to project root
**How to reproduce**

Start `nix develop` from inside of a sub-directory (not project root).

**Current behaviour**

`.pre-commit-config.yaml` gets added to _that sub-directory_

**Expected behaviour**

`.pre-commit-config.yaml` should always be added to project root (where `flake.nix` is) regardless of the $PWD from which `nix develop` was launched.

**Proposed solution**

Have an optional dependency on https://github.com/srid/flake-root and if the user has enabled it, have the devShell shellHook (of #266) do a:

```nix
shellHook = ''
  FLAKE_ROOT="''$(${lib.getExe config.flake-root.package})"
  (cd FLAKE_ROOT && ${config.pre-commit.installationScript})
''
```

srid:
Resolves #267 

Specifically implements: https://github.com/cachix/pre-commit-hooks.nix/issues/267#issuecomment-1490982847

The `FLAKE_ROOT` environment variable itself is a convention chosen by https://github.com/srid/flake-root

This PR solves the problem of `.pre-commit-config.yaml` being created willy-nilly in arbitrary subdirectories.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -387,22 +387,22 @@ in
 
             YAML_PATH="''${FLAKE_ROOT:-.}"/.pre-commit-config.yaml
 
-            if readlink ''$YAML_PATH >/dev/null \
-              && [[ $(readlink ''$YAML_PATH) == ${configFile} ]]; then
+            if readlink "''$YAML_PATH" >/dev/null \
+              && [[ $(readlink "''$YAML_PATH") == ${configFile} ]]; then
               echo 1>&2 "pre-commit-hooks.nix: hooks up to date"
             else
               echo 1>&2 "pre-commit-hooks.nix: updating $PWD repo"
 
-              [ -L ''$YAML_PATH ] && unlink ''$YAML_PATH
+              [ -L "''$YAML_PATH" ] && unlink "''$YAML_PATH"
 
-              if [ -e ''$YAML_PATH ]; then
+              if [ -e "''$YAML_PATH" ]; then
                 echo 1>&2 "pre-commit-hooks.nix: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml"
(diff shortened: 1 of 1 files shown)


srid:
An example of how this simplifies Nix when composing devShells,

https://github.com/nammayatri/nammayatri/pull/361/commits/c6ea11e0087c974b0be87a78ba19528006e29ac8

modules/pre-commit.nix
```diff
@@ -385,22 +385,24 @@ in
...
+            YAML_PATH="''${FLAKE_ROOT:-.}"/.pre-commit-config.yaml
+
+            if readlink ''$YAML_PATH >/dev/null \
+              && [[ $(readlink ''$YAML_PATH) == ${configFile} ]]; then
               echo 1>&2 "pre-commit-hooks.nix: hooks up to date"
             else
               echo 1>&2 "pre-commit-hooks.nix: updating $PWD repo"
 
-              [ -L .pre-commit-config.yaml ] && unlink .pre-commit-config.yaml
+              [ -L ''$YAML_PATH ] && unlink ''$YAML_PATH
```
roberth, line 396: These references should all be quoted in `"`.

roberth (commented):
Looks like an ok workaround that solves at least half of the problem. Ultimately though, a variable like this should be provided by `nix develop` itself.

modules/pre-commit.nix
roberth: including ones below

srid:
> Ultimately though, a variable like this should be provided by `nix develop` itself.

Agreed; https://github.com/NixOS/nix/issues/8145

srid:
Closing in favour of https://github.com/cachix/pre-commit-hooks.nix/pull/271

PR #271, opened 2023-04-06
Generate `.pre-commit-config.yaml` always at root of git working copy

### LINKED ISSUE #267: `.pre-commit-config.yaml` is not always installed to project root

srid:
Resolves #267

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -381,7 +381,7 @@ in
           elif ! ${git}/bin/git rev-parse --git-dir &> /dev/null; then
             echo 1>&2 "WARNING: pre-commit-hooks.nix: .git not found; skipping installation."
           else
-            GIT_WC=`${git}/bin/git rev-parse --git-dir`/..
+            GIT_WC=`${git}/bin/git rev-parse --show-toplevel`
 
             # These update procedures compare before they write, to avoid
             # filesystem churn. This improves performance with watch tools like lorri
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -393,7 +393,7 @@ in
             else
               echo 1>&2 "pre-commit-hooks.nix: updating $PWD repo"
 
-              [ -L "''${GIT_WC}/.pre-commit-config.yaml" ] && unlink "''${GIT_WC}/.pre-commit-config.yaml"
+              [ -L .pre-commit-config.yaml ] && unlink .pre-commit-config.yaml
 
               if [ -e "''${GIT_WC}/.pre-commit-config.yaml" ]; then
                 echo 1>&2 "pre-commit-hooks.nix: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml"
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -381,26 +381,28 @@ in
           elif ! ${git}/bin/git rev-parse --git-dir &> /dev/null; then
             echo 1>&2 "WARNING: pre-commit-hooks.nix: .git not found; skipping installation."
           else
+            GIT_WC=`${git}/bin/git rev-parse --git-dir`/..
+
             # These update procedures compare before they write, to avoid
             # filesystem churn. This improves performance with watch tools like lorri
             # and prevents installation loops by via lorri.
 
-            if readlink .pre-commit-config.yaml >/dev/null \
-              && [[ $(readlink .pre-commit-config.yaml) == ${configFile} ]]; then
+            if readlink "''${GIT_WC}/.pre-commit-config.yaml" >/dev/null \
+              && [[ $(readlink "''${GIT_WC}/.pre-commit-config.yaml") == ${configFile} ]]; then
               echo 1>&2 "pre-commit-hooks.nix: hooks up to date"
             else
               echo 1>&2 "pre-commit-hooks.nix: updating $PWD repo"
 
-              [ -L .pre-commit-config.yaml ] && unlink .pre-commit-config.yaml
+              [ -L "''${GIT_WC}/.pre-commit-config.yaml" ] && unlink "''${GIT_WC}/.pre-commit-config.yaml"
 
-              if [ -e .pre-commit-config.yaml ]; then
+              if [ -e "''${GIT_WC}/.pre-commit-config.yaml" ]; then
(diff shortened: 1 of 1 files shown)


srid:
@domenkozar - anything else I should do for this PR?

modules/pre-commit.nix
```diff
@@ -381,26 +381,28 @@ in
...
-              [ -L .pre-commit-config.yaml ] && unlink .pre-commit-config.yaml
+              [ -L "''${GIT_WC}/.pre-commit-config.yaml" ] && unlink "''${GIT_WC}/.pre-commit-config.yaml"
 
-              if [ -e .pre-commit-config.yaml ]; then
+              if [ -e "''${GIT_WC}/.pre-commit-config.yaml" ]; then
                 echo 1>&2 "pre-commit-hooks.nix: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml"
                 echo 1>&2 "    1. Translate .pre-commit-config.yaml contents to the new syntax in your Nix file"
                 echo 1>&2 "        see https://github.com/cachix/pre-commit-hooks.nix#getting-started"
                 echo 1>&2 "    2. remove .pre-commit-config.yaml"
                 echo 1>&2 "    3. add .pre-commit-config.yaml to .gitignore"
```
domenkozar, line 403: This step doesn't seem to be needed anymore.

modules/pre-commit.nix
```diff
@@ -381,26 +381,28 @@ in
...
 
-            if readlink .pre-commit-config.yaml >/dev/null \
-              && [[ $(readlink .pre-commit-config.yaml) == ${configFile} ]]; then
+            if readlink "''${GIT_WC}/.pre-commit-config.yaml" >/dev/null \
+              && [[ $(readlink "''${GIT_WC}/.pre-commit-config.yaml") == ${configFile} ]]; then
               echo 1>&2 "pre-commit-hooks.nix: hooks up to date"
             else
               echo 1>&2 "pre-commit-hooks.nix: updating $PWD repo"
 
-              [ -L .pre-commit-config.yaml ] && unlink .pre-commit-config.yaml
```
domenkozar, line 394: This should probably do what it did before, to unlink the file for a migration path.

modules/pre-commit.nix
```diff
@@ -381,26 +381,28 @@ in
...
-              if [ -e .pre-commit-config.yaml ]; then
+              if [ -e "''${GIT_WC}/.pre-commit-config.yaml" ]; then
                 echo 1>&2 "pre-commit-hooks.nix: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml"
                 echo 1>&2 "    1. Translate .pre-commit-config.yaml contents to the new syntax in your Nix file"
                 echo 1>&2 "        see https://github.com/cachix/pre-commit-hooks.nix#getting-started"
                 echo 1>&2 "    2. remove .pre-commit-config.yaml"
                 echo 1>&2 "    3. add .pre-commit-config.yaml to .gitignore"
               else
-                ln -s ${configFile} .pre-commit-config.yaml
+                ln -s ${configFile} "''${GIT_WC}/.pre-commit-config.yaml"
```
domenkozar, line 405:
```suggestion
                ln -fs ${configFile} "''${GIT_WC}/.pre-commit-config.yaml"
```

modules/pre-commit.nix
srid: Step 3, you mean? We still need to add to `.gitignore` because the YAML file is stored in project root. Or, are you okay with us putting it under `.git/` directory?

modules/pre-commit.nix
```diff
@@ -381,26 +381,28 @@ in
           elif ! ${git}/bin/git rev-parse --git-dir &> /dev/null; then
             echo 1>&2 "WARNING: pre-commit-hooks.nix: .git not found; skipping installation."
           else
+            GIT_WC=`${git}/bin/git rev-parse --git-dir`/..
```
roberth, line 384: This will put the file in `.git/worktrees` regardless of which worktree you're in.
If you're going the `--git-dir` route, you need to pit it inside the git dir. I think doing that was already suggested somewhere.

modules/pre-commit.nix
srid: > are you okay with us putting it under .git/ directory?

EDIT: I don't think we can do this, due to multiple worktrees (TIL) sharing the same `.git`: https://git-scm.com/docs/git-worktree

modules/pre-commit.nix
srid: Ah, so basically we want:

```sh
GIT_WC=`${git}/bin/git rev-parse --git-dir`
```

[Then](https://stackoverflow.com/a/75780328/55246), the YAML file would be put under `.git/worktrees/foo` - and thus we can obviate having to gitignore it.

roberth: Different worktrees may use different revisions of the hooks, leading to a branch to be reformatted with the wrong logic.

srid: Besides, `.git` can also exist completely outside of the (only) working tree. So the `/..` appending is wrong.

So, if I understand you correctly, `GIT_WC=`${git}/bin/git rev-parse --git-dir`` is where we should put the YAML. That is guaranteed to return a path unique to that worktree.

roberth: I don't think `pre-commit --config` is clever enough to work with the right `--git-dir` though. It appears as a relative path that's hardcoded in the hook script. And there's only one hook script.

It doesn't have shell injection like `pre-commit install --config='.git/pre-commit-hooks.yaml'`, and it's not clever about `.git/`.

roberth: Maybe just go with `git rev-parse --show-toplevel`?
That seems to have been your original intent and seems to work fine.

srid: Alright, done.

PR #274, opened 2023-04-11
Add hook for `checkmake`

### LINKED ISSUE #251: Add checkmake
Please add [checkmake](https://github.com/mrtazz/checkmake), a linter for Makefiles.

* `checkmake` is available on Nixpkgs
* The check command is `checkmake <makefile>`

Niols:
closes #251

This PR adds a hook for [`checkmake`](https://github.com/mrtazz/checkmake), an “experimental linter/analyzer for Makefiles.”


Niols:
Turns out `checkmake` is not supported on `x86_64-darwin` (at least). Is there a solution to handle this situation?

domenkozar:
We currently only have tools that also work on darwin, the best way is to fix darwin support.

Niols:
Turns out there is a new version of `checkmake` and it seems to build just fine on Darwin. It is going to land into nixpkgs soon-ish hopefully. cf https://github.com/NixOS/nixpkgs/pull/225738

What would be the process to include a tool that is available in nixpkgs unstable, but only available in nixpkgs stable in Linux but not Darwin?

Niols:
https://github.com/NixOS/nixpkgs/pull/225738 now landed. At the next `nixos-unstable` snapshot, `checkmake` 0.2.2 should then be available for both Linux and Darwin and this PR should start working fine. I will look for a null-trick-inspired way of making the flake checks for `nixpkgs-stable` happy on Darwin.

Niols:
3b2fd74 is not mandatory, but it shows that things work rather fine both on Ubuntu and MacOS. Of course, this is still problematic in `nixpkgs-stable`.

Niols:
OK, found a way that I believe is quite clean. WDYT? I can also remove the commit that updates the `nixpkgs` input; I only included it to show that tests passed once `nixpkgs` (unstable) was recent enough.

PR #292, opened 2023-04-27
Adding support for Elixir pre-commits

atavistock:


modules/hooks.nix
```diff
@@ -1223,5 +1233,35 @@ in
...
+      mix-format = {
+        name = "mix-format";
+        descriptiion = "Runs the built-in Elixir syntax formatter";
+        entry = "${pkgs.elixir}/bin/mix format";
+        types = [ "elixir" ];
+      };
+
+      mix-test = {
+        name = "mix-test";
+        descriptiion = "Runs the built-in Elixir test framework";
```
domenkozar, line 1246: There's a typo in `descriptiion` in all definitions

domenkozar:
@atavistock do you want to fix this PR?

atavistock:
Yes, just haven't been able to circle back to it.  Will have more time
later this week.

atavistock:
@domenkozar Fixed the typos and updated to match current `master` branch

atavistock:
Looks like a close bracket was lost, now fixed.

domenkozar:
You'll need to rebase on top of master.

domenkozar:
Ping :)

nifoc:
As far as I can tell, `pre-commit` doesn't understand `elixir` as a type tag. I get the following error:

```
Type tag 'elixir' is not recognized.  Try upgrading identify and pre-commit?
```

Checking the `identify` repository, it indeed looks like `elixir` is not supported at the moment.

How is this supposed to work? Can you somehow configure type tags manually? (Couldn't find anything obvious on [pre-commit.com](https://pre-commit.com))

nifoc:
Submitted a PR to `identify` and the above error is now gone.

However, `nix flake check` does not work, because:

- Dependencies are not copied into the drv (because they're part of the `.gitiginore`)

```
       > Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :ecto from Pan.MixProject
       > Could not find Hex, which is needed to build dependency :ecto
```

PR #301, opened 2023-05-18
deadnix: Use mkCmdArgs, expose more args

mweinelt:
Make the argument handling less arcane and instead use the mkCmdArgs standard facility.

Exposes the new parameters `--hidden` to scan hidden dirs/files and `--exclude` to exclude individual paths.

cc @cyntheticfox

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -64,6 +64,20 @@ in
               default = false;
             };
 
+          exclude =
+            mkOption {
+              type = types.listOf types.str;
+              description = lib.mdDoc "Files to exclude from analysis.";
+              default = [ ];
+            };
+
+          hidden =
+            mkOption {
+              type = types.bool;
+              description = lib.mdDoc "Recurse into hidden subdirectories and process hidden .*.nix files.";
+              default = false;
+            };
+
           noLambdaArg =
             mkOption {
               type = types.bool;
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -553,10 +567,18 @@ in
...
-              args = lib.concatMapStringsSep " " toArg (lib.filter (attr: settings.deadnix."${attr}") (lib.attrNames settings.deadnix));
+              cmdArgs =
+                mkCmdArgs (with settings.deadnix; [
+                  [ noLambdaArg "--no-lamda-arg" ]
+                  [ noLambdaPatternNames "--no-lambda-pattern-names" ]
+                  [ noUnderscore "--no-underscore" ]
+                  [ quiet "--quiet" ]
+                  [ hidden "--hidden" ]
+                  [ edit "--edit" ]
+                  [ (exclude != []) "--exclude ${concatStringsSep " -- " exclude}"]
```
cyntheticfox, line 578: The purpose of having the `--` below was as the argument-list terminator (completely unnecessarily). Concatenating the excludes like this probably won't parse correctly given deadnix's [clap macro call](https://github.com/astro/deadnix/blob/main/src/main.rs) seems to expect you to pass the flag for each different value (not supporting a comma-separated-list like I expected), which seems weird to me (`clap` `ArgAction::get_many()`)

cyntheticfox (commented):
The changes to cmdArgs looks good, but I'm not sure if the excludes are passed correctly for how `clap` parses them

modules/hooks.nix
mweinelt: > The --exclude parameter accepts multiple paths. Separate them with -- to pass [FILE_PATHS]....

In understood this as one `--exclude` parameter accepts multiple paths separated by `--`. Are you saying I'm misunderstanding that or that it is misleading?

PR #309, opened 2023-05-29
nil: Fix hook only failing if last file fails

Enzime:
As `set -e` is not set, the script only inherits the exit code of the last command which would be `nil diagnostics $file` on the last file in the array.

As `pre-commit` separates the files into separate batches and invokes the script multiple times, I think that it's better for the script to not exit early so that `nil` can be run on all the files.

Adding messaging on stdout is also useful for debugging, by default, it only appears when the hook fails, or when someone specifies `-v` to `pre-commit run`.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -731,9 +731,20 @@ in
           entry =
             let
               script = pkgs.writeShellScript "precommit-nil" ''
+                errors=false
+                echo Checking: $@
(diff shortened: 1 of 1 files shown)


domenkozar:
Would it be simpler to `set -e`?

Enzime:
Definitely simpler, but I think the benefit to the user of checking all the files outweighs the complexity of this script

PR #312, opened 2023-06-04
bump fourmolu (0.9 -> 0.12), nixos-23.05

smunix:
- bump fourmolu version from 0.9 to 0.12
- bump to nixos-23.05, nixpkgs-unstable

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -17,12 +17,15 @@
       }
     },
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1667395993,
-        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "lastModified": 1685518550,
+        "narHash": "sha256-o2d0KcvaXzTrPRIo0kOLV0/QXHhDQ5DTi+OxcjO8xqY=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "5aed5285a952e0b949eb3ba02c12fa4fcfef535f",
+        "rev": "a1720a10a6cfe8234c0e93907ffe81be440f4cef",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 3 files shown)


smunix:
@domenkozar how can I get the action checks to trigger  on this PR?

maralorn:
Sorry, if I remarked this already somewhere else (then I can’t find it again). Just wanted to note that there are three arguments against this PR, which caused me friction:

* It breaks caching when just used with cache.nixos.org
* It leads to inconsistent formatting between the pre-commit hook and fourmolu installed from shell.
* It tightly couples this repo to a nixpkgs commit. When used together with current nixos-unstable it throws an eval error.

domenkozar:
Happy to revert, I agree.

smunix:
> ... fourmolu installed from the shell 

Is this related to `devenv` shell? If yes, then wouldn't bumping pre-commit in devenv be a better fix? 

The older fourmolu doesn't compile with recent ghc96, so I'd advocate against reverting.

@maralorn , how can I reproduce the issue you're facing with this? 

@domenkozar I can push a patch to `caxhix/devenv` to bump `pre-commit`, but I'm afraid I don't entirely understand what the issue is?

maralorn:
@smunix This is not related to devenv. The problem is that a completely normal mkShell with haskellPackages.fourmolu will contain another version of fourmolu than the pre-commit hook, which is annoying because they do not format in the same way.

In my opinion the best solution if we want to use a newer fourmolu, is to bump the default fourmolu in nixpkgs to a newer version, which is actually something which I generally regard as a good idea. (And I am the fourmolu maintainer in nixpkgs.) Especially because hls already uses a newer fourmolu version and we really should be using consistent versions across all setups.

> The older fourmolu doesn't compile with recent ghc96, so I'd advocate against reverting.

The default compiler version of nixos-unstable is ghc 9.2. I don’t see why you need a bleeding edge ghc to compile your pre-commit hooks. But granted that in your larger project context you use ghc 9.6 for everything and want a fourmolu version which works with that, then you can override fourmolu for the pre-commit hook with

```
settings.tools.fourmolu = lib.mkForce pkgs.haskell.packages.ghc96.fourmolu;
```

You can reproduce my current problem by setting the `pkgs` variable passed to pre-commit nix to a nixos-unstable newer than Yesterday. Then you get this error:
```
       error: attribute 'fourmolu_0_12_0_0' missing

       at /nix/store/2wdwg4ib0c4m3vpggg8rv5kwflgr7qwf-source/nix/tools.nix:118:14:

          117|   # TODO: these two should be statically compiled
          118|   fourmolu = haskellPackages.fourmolu_0_12_0_0;
             |              ^
          119|   inherit (luaPackages) luacheck;
       Did you mean one of fourmolu_0_13_0_0, fourmolu_0_10_1_0 or fourmolu_0_3_0_0?
```

Sadly the above given workaround to picking a different formolu does not prevent the missing binding to be evaluated, so there is no downstream way to fix this.

smunix:
Ah, I see. Specifying that `fourmulu_<version>` is prone to breaking as the upstream nixpkgs moves.

> In my opinion the best solution if we want to use a newer fourmolu, is to bump the default fourmolu in nixpkgs to a newer version, which is actually something which I generally regard as a good idea

☝️ Agreed! Let's revert this PR. Thanks for catching and working on this

smunix:
> settings.tools.fourmolu = ...

@domenkozar , is it possible to override this ☝️ in `devenv`? It would be great to have indeed

PR #330, opened 2023-07-27
Make the generated configuration depend on the pre-commit package

K900:
Fixes the hooks not updating when the pre-commit package changes but the config doesn't.


domenkozar:
Needs to run the hooks :)

K900:
That's what I get for using the web UI.

Issue #336, opened 2023-08-08
pre-commit config references a broken symlink in nix store


alejandro-angulo:
**Describe the bug**
When trying to set `pre-commit.settings.prettier.binPath` my generated `.pre-commit-config.yaml` file ends up with an entry referencing a broken symlink in my nix store.

Here's an example of generated config file https://gist.github.com/alejandro-angulo/a467575f910ea4bb7b837d913f8e44af

**To reproduce**
I've created a repo to reproduce this bug: https://github.com/alejandro-angulo/devenv-pre-commit-test/tree/59e2e768d472d64e39e9d47992fd54c83cfb9d31

```
git clone https://github.com/alejandro-angulo/devenv-pre-commit-test.git
cd devenv-pre-commit-test
nix develop --impure  # or `direnv allow`
pre-commit run prettier --files test.js
```
On my machine I get this output
```
❯ pre-commit run prettier --files test.js
prettier.................................................................Failed
- hook id: prettier
- exit code: 1

Executable `/nix/store/apjv7ccblhywdccadbpb2w0ghd0k20fb-prettier` not found

❯ l /nix/store/apjv7ccblhywdccadbpb2w0ghd0k20fb-prettier
lrwxrwxrwx 28 root 31 Dec  1969  /nix/store/apjv7ccblhywdccadbpb2w0ghd0k20fb-prettier -> ../prettier/bin/prettier.cjs
```

Notice that the symlink resolves to a directory `/nix/store/prettier/....` (this does not exist).

I expected the executable to match the value I see in `nix repl`

```
nix-repl> outputs.devShells.x86_64-linux.default.config.pre-commit.settings.prettier.binPath
/nix/store/a80wmqkzayj1kvrllif6db3c0skklymr-source/node_modules/.bin/prettier
```

**Version**
devenv: 0.6.3

domenkozar:
I transfered it to the right repo. Thanks!

alejandro-angulo:
Oh I found out that I can instead have `pre-commit.settings.prettier.binPath = self.outPath + "/node_modules/.bin/prettier"` and it works as I expected.

PR #337, opened 2023-08-10
Added skipBinCopy option for prettier and eslint

alejandro-angulo:
This makes it so that the logic agrees with the docs: https://devenv.sh/reference/options/#pre-commitsettingsprettierbinpath . If you disagree with the spirit of this change then I think at the least the documentation should be updated because what it suggests today doesn't work.

Relevant issue is #336 , I've added a branch named `with-patched-pre-commit-hooks-nix` to the repo I used to reproduce the issue and it seems to work. Relevant change in the repo: https://github.com/alejandro-angulo/devenv-pre-commit-test/compare/main...with-patched-pre-commit-hooks-nix .

This is basically the same as what I suggested in the issue earlier but I decided to name the option `skipBinCopy` rather than `binPathInRepo`.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1194,6 +1194,7 @@ in
           name = "prettier";
           description = "Opinionated multi-language code formatter.";
           entry = with settings.prettier;
+            # Casting `binPath` to a string will prevent the path from being copied to the nix store.
             "${if skipBinCopy then toString binPath else binPath} ${lib.optionalString write "--write"} ${lib.optionalString (output != null) "--${output}"} --ignore-unknown";
           types = [ "text" ];
         };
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -219,13 +219,21 @@ in
...
               '';
             };
 
+          skipBinCopy =
+            mkOption {
+              type = types.bool;
+              description = lib.mdDoc
+                "If set, prevents the binary from being copied to the nix store. This should only be set if the binary is already committed to your repo.";
+              default = false;
+            };
```
alejandro-angulo, line 235: Not sure I like introducing this option but maybe I'm overthinking it :shrug:

alejandro-angulo:
Ah wait I figured out I could accomplish this without making any changes by setting `pre-commit.settings.eslint.binPath = self.outPath + "/node_modules/.bin/prettier";`

alejandro-angulo:
Ah wait I figured out I could accomplish this without making any changes by setting `pre-commit.settings.eslint.binPath = self.outPath + "/node_modules/.bin/prettier";`

johnhamelink:
@alejandro-angulo's solution doesn't work because the `node_modules` directory is usually kept out of git, and so is not part of the outPath.

sandydoo:
To use local `node_modules`, you can pass in a string instead of a path. You'll need to use an impure mode, like `nix fmt` or `nix develop --impure`. This has been addressed in the docs for `binPath`.

PR #346, opened 2023-09-02
fix(stylua): don't pass in files so it picks up .styluaignore

mrcjkb:
When passing in the file name to `stylua`, it bypasses [`.styluaignore`](https://github.com/JohnnyMorganz/StyLua#filtering-using-styluaignore).

See also: https://github.com/JohnnyMorganz/StyLua/issues/751

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1095,6 +1095,8 @@ in
           description = "An Opinionated Lua Code Formatter.";
           types = [ "file" "lua" ];
           entry = "${tools.stylua}/bin/stylua";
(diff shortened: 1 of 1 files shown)


mrcjkb:
Converting this to draft, because it looks like stylua might respect .styluaignore in the future.

domenkozar:
What should we do here?

mrcjkb:
> What should we do here?

I'm currently not blocked by this.
We could close this PR and I'll open a new one when stylua gets its `--respect-ignores` flag.

PR #351, opened 2023-09-10
Add pre-commit hook for lychee (fast link checker)

totoroot:
[lychee](https://github.com/lycheeverse/lychee/) is a much more performant linkchecker than `mkdocs-linkcheck`, which is currently the only linkchecking tool available in pre-commit-hooks.nix.
It provides a couple additional, useful features, is much faster, and the closure size is currently about ~21MB IIRC.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -722,6 +722,22 @@ in
             default = "";
           };
       };
+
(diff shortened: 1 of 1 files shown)


domenkozar:
Could you rebase?

Issue #219, opened 2022-12-31
alejandra: provide a `package` option


NobbZ:
Currently some `alejandra` is used, and for the user it is not clear from where it is taken.

Please document how the default `alejandra` is picked and give an option to easily override it.

NobbZ:
Similar restrictions seem to apply for all other hooks.

Can't currently use the rust based ones as they seem to use rust from `nixpkgs`, while I am developing using a nightly rust compiler…

domenkozar:
See https://github.com/cachix/devenv/blob/main/src/modules/languages/rust.nix#L59 how overrides can be done

purefn:
I'm currently trying this with 

```
  pre-commit-check = nix-pre-commit-hooks.run {
    src = ./.;
    excludes = ["^vendor/"];
    hooks = {
      ormolu.enable = true;
      purs-tidy.enable = true;
    };
    settings = {
      ormolu.defaultExtensions = [ "TypeApplications" "BangPatterns" "QuasiQuotes" "TemplateHaskell" ];
    };
    tools =
      let
        mkOverrideTools = lib.mkOverride (lib.modules.defaultOverridePriority - 1);
      in {
        purs-tidy = mkOverrideTools purs-tidy;
        ormolu = mkOverrideTools haskell-pkgs.ormolu;
      };
  };
```

to get the version of ormolu that we've been using and it doesn't seem to be working as expected. What am I missing?

purefn:
Ignore my comment. It was using the version I was telling it. It was elsewhere that wasn't using the version I was expecting. Sorry!

Issue #218, opened 2022-12-31
alejandra: use `--check`


NobbZ:
The current implementation of the `alejandra` hook runs `alejandra` and changes the files on the filesystem, somehow checking for dirty/changed files after that.

I do not like when git hooks actually alter the repo, they should simply signal "failure" or "success" through the exit code.

To solve this issue one can use `--check` with `alejandra` though there is no option exposed to enable that.

Please make `--check` the default and/or add an option to opt-in/out.

totoroot:
I did not change the defaults, but introduced a `check` option that can be used to achieve what you were trying to do.
More information on that can be found in #353.

PR #347, opened 2023-09-02
Add alejandra verbosity settings

adam-gaia:
Hello,
This PR adds two settings to `alejandra` that control verbosity flags.
- quiet (bool): maps to `alejandra --quiet`
- silent (bool): maps to `alejandra --quiet --quiet`. While it look strange, the double argument is intentional. It makes a little more sense with the short format, `-qq`, but I spelled out the long version for better readability.

The `quiet` option is redundant when `silent` is set. An enum with normal/quiet/silent options would have been ideal here. I'm still learning the nix language and couldn't come up with a better way to do this. If you have any thoughts, let me know!

PR #353, opened 2023-09-10
Add args to alejandra hook

totoroot:
I introduced settings for all available command line arguments for the existing `alejandra` hook.

This allows for specifying `settings.alejandra.check = true;` to only check the style of files and not writing changes to them. This was requested by @Nobbz in #218. I did not change the default behaviour of the hook, so without the `check` setting, it still writes changes.

It also adds a `package` option as was requested in #219.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -80,12 +80,47 @@ in
         };
       alejandra =
         {
+          check =
+            mkOption {
+              type = types.bool;
+              description = lib.mdDoc "Check if the input is already formatted and disable writing in-place the modified content";
+              default = false;
+              example = true;
+            };
           exclude =
             mkOption {
               type = types.listOf types.str;
               description = lib.mdDoc "Files or directories to exclude from formatting.";
               default = [ ];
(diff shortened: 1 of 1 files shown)


totoroot:
I only just noticed that @adam-gaia's PR #347 did part of what I did here. Sorry for that, but I hope that they wouldn't mind if this gets merged instead of their PR, as I introduced more options and also handled the issue with both `quiet` and `silent` being set.

totoroot:
I liked the suggestion of @adam-gaia to use `types.enum` for the verbosity setting, as the way `alejandra` handles verbosity is somewhat unconventional.
I added another commit that hopefully also fixed the build failure.

adam-gaia:
> I only just noticed that @adam-gaia's PR #347 did part of what I did here. Sorry for that, but I hope that they wouldn't mind if this gets merged instead of their PR, as I introduced more options and also handled the issue with both `quiet` and `silent` being set.

@totoroot I don't mind at all. Thanks for doing this the right way :)

PR #361, opened 2023-10-05
Fix broken import (eclint)

totoroot:
The merge of PR #359 caused pipelines to fail as master is now broken due to a missing import.

Since the import is not strictly needed, I'd say it is better to remove it again.

Also, notice that `eclint` in nixpkgs is not the [archived eclint](https://github.com/jednano/eclint) written in JS, but a [maintained alternative also named eclint](https://gitlab.com/greut/eclint) written in Go. Thus, if listing it in `nix/tools.nix` it does not belong in `nodePackages`.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -304,13 +304,13 @@ in
         };
       eclint =
         {
-          binPath =
+          package =
             mkOption {
-              type = types.path;
-              description = lib.mdDoc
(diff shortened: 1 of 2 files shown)


domenkozar:
It should come from tools instead of pkgs, for sale of consistency

totoroot:
Ok, I'll fix it and add some arguments as well.

PR #364, opened 2023-10-09
Switch type to one supported by precommit

atavistock:
Even though python `identify` registers these with a tag of `elixir`, it looks like `precommit` only supports some tags as types.  This change converts these to supported types.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1238,14 +1238,14 @@ in
         name = "mix-format";
         descriptiion = "Runs the built-in Elixir syntax formatter";
         entry = "${pkgs.elixir}/bin/mix format";
-        types = [ "elixir" ];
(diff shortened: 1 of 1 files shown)


akirak:
> Even though python identify registers these with a tag of elixir, it looks like precommit only supports some tags as types. 

This is because the stable channel of nixpkgs hasn't received the recent version of pre-commit yet. If you feed the unstable channel of nixpkgs to pre-commit-hooks, `elixir` tag will be handled:

```nix
  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  inputs.pre-commit-hooks = {
    url = "github:cachix/pre-commit-hooks.nix";
    inputs.nixpkgs.follows = "nixpkgs";
  };
```

PR #365, opened 2023-10-19
Specify files attribute for Elixir programs

akirak:
After #364, the Elixir hooks run on other syntax's too (e.g. Nix), which is not an expected behavior. `elixir` type will be supported on stable soon, but this PR resolves the issue for now.

The regular expression is based on this commit: https://github.com/pre-commit/identify/commit/3b0128c58d4b401f1a937cb79fadc2e2594f9ce7


domenkozar:
cc @atavistock

PR #369, opened 2023-11-04
Add git-annex pre-commit hook

poelzi:
Adds git-annex pre-commit hook which is required, when you want to use git-annex in a repository.
git-annex adds large file support and automatic sync to git

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,36 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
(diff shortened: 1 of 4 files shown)


domenkozar:
Could you remove the editor config?

poelzi:
Having a .editorconfig helps getting the style correct in the first place without running through formatting. My IDE also shows intention as red if they are not correct.
It is a [standard](https://editorconfig.org/) supported by many editors. What harm does it do ?

domenkozar:
This hook is failing on darwin, can you take a look?

poelzi:
It seems that git-annex is broken on darwin. I created ticket on [git-annex.branchable.com](https://git-annex.branchable.com/bugs/test__58___posix__95__spawnp_broken_10_on_darwin)

PR #379, opened 2023-12-21
feat: add typstfmt

storopoli:
- adds typstfmt (https://github.com/astrale-sharp/typstfmt)
- most of the settings were ported from https://github.com/astrale-sharp/typstfmt/blob/master/.pre-commit-hooks.yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,10 @@ use nix
 
 - [taplo fmt](https://github.com/tamasfe/taplo)
 
+## Typst
+
+- [typstfmt](https://github.com/astrale-sharp/typstfmt)
+
 ## Fortran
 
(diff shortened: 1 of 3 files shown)


storopoli:
I don't understand the CI errors. It is suggesting `typst-fmt` whereas all the changes are using `typstfmt` and nix pkg is called `typstfmt`: https://search.nixos.org/packages?type=packages&query=typstfmt

domenkozar:
It seems that in 23.05 it's called typst-fmt

storopoli:
Ok, I've changed to `typst-fmt` we might need to change to `typstfmt` once 23.11 comes.

PR #387, opened 2024-01-11
typos: unset pass_filenames

phip1611:
Typos is supposed to run on the whole tree. If this is set to true, the system gets stuck for large projects due to very high memory consumption. The restriction on with files typos run, should be specified in the typos config file.

Tested this locally with a large project.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -892,6 +892,7 @@ in
               type = types.str;
               description = lib.mdDoc "Path to a custom config file.";
               default = "";
+              example = ".typos.toml";
             };
 
           diff =
@@ -2153,6 +2154,11 @@ in
             in
             "${tools.typos}/bin/typos ${cmdArgs}${lib.optionalString settings.typos.diff " --diff"}${lib.optionalString settings.typos.hidden " --hidden"}";
           types = [ "text" ];
+          # Typos is supposed to run on the whole tree. If this is set to true,
+          # the system gets stuck for large projects due to very high memory
+          # consumption. The restriction on with files typos run, should be
+          # specified in the typos config file.
+          pass_filenames = false;
         };
       typstfmt = {
         name = "typstfmt";


domenkozar:
Could we link to some upstream issue about memory consumption?

phip1611:
Relevant upstream issues/PRs are:
- https://github.com/crate-ci/typos/issues/738
- https://github.com/crate-ci/typos/pull/880
- https://github.com/crate-ci/typos/pull/791

TL;DR: `typos` is `26MiB` in size. In a large project of mine (thousand of files), my system always freezes without `pass_filenames = false`. Unfortunately, it was not possible to see how the `typos` utility is invoked in the end, and `--verbose` doesn't print that.

I'm not entirely sure, and didn't investigate it with `strace` or so.

Other tools that are supposed to run on the tree such as `rustfmt` use the same setting.

phip1611:
Update. I used strace and knows what happens. 

So again, the project is very huge. If I run `$ typos .`, it takes a couple of seconds, but that's it. When pre-commit runs `typos` with `pass_filenames = true`,  `pre-commit` wants to pass all `74000` files of the project to `typos`. I guess there is some limit in pre-commit hooks so that it splits these huge amount of arguments to 8 instances of `typos`.

I suppose a combination of the relatively big memory usage of `typos` in combination with the ridiculous amount of parameters  passed to it, results in an enormous amount of memory usage.

TL;DR: It is correct to use `pass_filenames = false`,  as `typos` is supposed to run on the whole tree and excludes are specified in `.typos.toml`. This aligns the behavior of `typos` via `$ pre-commit run` with `$ typos .`, if the right config file is passed.

totoroot:
@phip1611 I don't think this change is desirable. While I understand the issue at hand, changing the default behaviour of the hook to align with the call of `typos .` is not a good idea IMO.

I disagree with your statement "Typos is supposed to run on the whole tree.". pre-commit hooks are supposed to be run on changes that are being committed, not for the whole repository every time you commit changes.

If you wished to run the pre-commit hook on the whole directory/repository, you always had the option to run `pre-commit run typos --all`.

The way the hook is set up now, there is no option to just check a single file that you'd like to commit and ignore the rest of the repository.

This change 'broke' the `typos` hook in one of our repositories, as now all files get checked, even when just wanting to commit a single new file. This repository had hundreds of files with typos in them and while it is planned to check and fix them eventually, the hook should not care about any files beside the one that I'm trying to commit.

After all, it is a **pre-commit** hook.

@domenkozar I'd suggest that this gets reverted and we discuss how we can otherwise fix the issue that @phip1611 was running into.

phip1611:
In the meantime, I had more time to better understand the issue and thought about it again. 

> Were you committing the whole project at once?

No. But I want to be able to run `pre-commit run --all-files`. In that case, `typos` should run the same as `$ typos . --config .typos.toml`. But I think we can combine both worlds.

We can set `hooks.typos.excludes (pseudo code):

```
excludes =
          let
            configFile = builtins.readFile ../.typos.toml;
            config = builtins.fromTOML configFile;
          in
          config.files.extend-exclude;
```

Stunkymonkey:
I can report this PR breaks my config having:
```
pre-commit.settings.hooks.typos = {
    enable = true;
    excludes = [ "secrets\\.yaml" "\\.sops\\.yaml" ];
};
 ```

sure I can workaround with:

```
settings.typos.exclude = "(secrets|\.sops)\.yaml";
```

But the `excludes` is always ignored. We should revert the change with `pass_filenames`. 

@totoroot said correctly: 

> A pre-commit hook is supposed to only check files that are about to be committed and not the whole repository.

PR #388, opened 2024-01-12
Add args to prettier hook

totoroot:
Add pretty much all available options/command line arguments to the `prettier` hook.

See all CLI flags for prettier [here](https://prettier.io/docs/en/cli.html).
See all options for prettier [here](https://prettier.io/docs/en/options.html).

That both `--write` and `--ignore-unknown` were set to true by default, deviates from prettier's default behaviour.
Especially defaulting to overwriting files by setting `--write` seems unwanted for a pre-commit hook so the default is now set to false.

`--ignore-unknown` can be set via option now, but also defaults to false.

The `output` option was removed in favour of having `check` and `list-different` as separate options.
`check` defaults to `true`, so that prettier can be used sensibly as a pre-commit hook.

If a user prefers the output format of `list-different`, they now have the option to enable it.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -687,29 +687,244 @@ in
               '';
             };
         };
+      # See all CLI flags for prettier [here](https://prettier.io/docs/en/cli.html).
+      # See all options for prettier [here](https://prettier.io/docs/en/options.html).
       prettier =
         {
           binPath =
             mkOption {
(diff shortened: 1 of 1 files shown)


totoroot:
Added a patch file by accident. Removed it.

All introduced options were thoroughly tested.
Ready for review.

PR #391, opened 2024-01-13
fix: fix up tools indirections, ensure the used tools actually exist

K900:


K900:
Fixes a bunch of hooks failing after #389

domenkozar:
CI fails

K900:
Well that's interesting. NixOS 23.05 did not have the attributes the hooks referred to, so now they don't evaluate. NixOS 23.11 has a different problem in that terraform is now unfree, so the hooks don't evaluate again.

K900:
All right, I think it should be good now.

K900:
Added a test to make sure this doesn't regress again. Unfortunately there's still the OSX build failure...

domenkozar:
Great, let's fix this as a follow up :)

totoroot (commented):
Thanks for this!

PR #392, opened 2024-01-16
Extend options for `typos` hook

totoroot:
### Adding options for existing flags
I've added a few more options to the `typos` hook to pass existing command line flags.
These options are `binary`, `no-check-filenames`, `no-check-files`, `no-unicode`, `quiet`, `verbose`.
I also refactored the descriptions of the options and `cmdArgs` of the hook a bit.

### Adding an option to ignore files
[I asked in the `typos` repo](https://github.com/crate-ci/typos/discussions/907) how to best ignore words, that should not be marked as typos. The way `typos` works, you have to add these words/spellings to the word list by extending it.
Due to some design decisions, `typos` maintainers will not introduce CLI flags to ignore words and will to stick to the use of a single config file (_typos.toml) for that.

To make `typos` even more usable as a pre-commit hook, I've added an option to list words to be ignored in the configuration of the hook. This makes it possible to set the following:

```nix
settings.typos.ignored-words = [
  "mqtt"
  "mosquitto"
  ];
```

If the option is set, the list of words then gets appended to a passed config file in a way `typos` understands.
The resulting config file passed to the hook then contains the required section:

```toml
[default.extend-words]
mqtt = "mqtt"
mosquitto = "mosquitto"
```

If both `settings.typos.config` and `settings.typos.configPath` are set, the latter gets ignored. This is now documented in the option's description, so that users know not to configure both.

### Unsetting `pass_filenames`
In commit f441d0892f47d199a1a96b0c0efec13fdbff01ee, I have removed the recent change from @phip1611. [I stated in the PR discussion](https://github.com/cachix/pre-commit-hooks.nix/pull/387#issuecomment-1893321739) why I think changing this was not a good idea.
@domenkozar Perhaps we can wait for @phip1611's response and discuss this in https://github.com/cachix/pre-commit-hooks.nix/pull/387 before this PR gets merged.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1087,17 +1087,22 @@ in
         };
       typos =
         {
+          binary =
+            mkOption {
+              type = types.bool;
+              description = lib.mdDoc "Whether to search binary files.";
+              default = false;
+            };
           color =
             mkOption {
               type = types.enum [ "auto" "always" "never" ];
               description = lib.mdDoc "When to use generate output.";
               default = "auto";
             };
-
           config =
             mkOption {
               type = types.str;
-              description = lib.mdDoc "Multiline-string configuration passed as config file.";
+              description = lib.mdDoc "Multiline-string configuration passed as config file. If set, config set in `typos.settings.configPath` gets ignored.";
               default = "";
               example = ''
                 [files]
@@ -1110,7 +1115,6 @@ in
                 extend-glob = []
               '';
             };
-
           configPath =
             mkOption {
               type = types.str;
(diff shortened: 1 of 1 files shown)


phip1611:
Please consider adding this patch:

https://github.com/cachix/pre-commit-hooks.nix/pull/387#issuecomment-1893600631

Discussion is in the issue #387.

phip1611:
@totoroot any update? I'm only available for one more week before I'm on vacation; I hope we can merge this soon :)

domenkozar:
Ping :)

totoroot:
Sorry, for the late response. I missed the notifications on this one.

@phip1611 I tried incorporating your patch, but did not see how this fixes the issues that only came up after https://github.com/cachix/pre-commit-hooks.nix/pull/387 got merged.
As @Stunkymonkey has also reported this change breaking their config in #387, I'd still suggest reverting the change by merging commit https://github.com/cachix/pre-commit-hooks.nix/pull/392/commits/f441d0892f47d199a1a96b0c0efec13fdbff01ee as par of this PR.
Then you can open up a follow-up PR with your proposed patch and we can thoroughly test it and see whether the hook then works for you, @Stunkymonkey and me.

@domenkozar Is that alright?

PR #398, opened 2024-02-16
Extend `yamllint` hook

totoroot:
I added a couple of options to the `yamllint` hook, as so far it was only possible to pass a path to the configuration file.
In some comments, I also reasoned why some options weren't added as they do not make sense for a pre-commit hook.

Similar to how I did it for the `typos` hook, there is now also an option to pass a multiline string as configuration to `yamllint`. Additionally `yamllint` allows for a serialized configuration, but the multiline string configuration option is the preferred way. Of course just providing a path to a configuration file remains possible.

The behaviour of the hook itself mostly stayed the same, although the `relaxed` option was removed in favour of a more generalized `preset` option, that allows us to easily add presets, should more ever be added to `yamllint`.

A pre-commit hook only makes sense, if it prevents you from committing if errors were found, so `--strict` should definitely be the default.

### Breaking changes:
- `settings.yamllint.relaxed = true;` is now `settings.yamllint.preset = "relaxed";`
- `settings.yamllint.strict` was added and defaults to `true` :arrow_right: strict error checking for hook

### Additional changes:
For some hooks I recently worked on, I noticed that I was shadowing the top-level attribute `config`, which is never a good idea, so I changed the multiline configuration options to `configuration`.
Affected hooks are `typos` and `vale`. The options were only added recently.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -2528,7 +2528,7 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
         let
           # .luarc.json has to be in a directory,
           # or lua-language-server will hang forever.
-          luarc = pkgs.writeText ".luarc.json" (builtins.toJSON hooks.lua-ls.settings.config);
+          luarc = pkgs.writeText ".luarc.json" (builtins.toJSON hooks.lua-ls.settings.configuration);
           luarc-dir = pkgs.stdenv.mkDerivation {
             name = "luarc";
             unpackPhase = "true";
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -586,7 +586,7 @@ in
                 "The diagnostic check level";
               default = "Warning";
             };
-            config = mkOption {
+            configuration = mkOption {
               type = types.attrs;
               description = lib.mdDoc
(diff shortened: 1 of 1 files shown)


domenkozar:

modules/hooks.nix
```diff
@@ -1340,7 +1340,7 @@ in
                 default = "auto";
               };
 
-            config =
+            configuration =
```
domenkozar, line 1343: we should warn about this option rename and deprecate `relaxed`

totoroot: Sure. How do I create a warning?

totoroot: I tried adding them in 2602c35e7dca1e4d4c32b8f4b308e555948069ba but it fails with the following:
```nix
error: evaluation aborted with the following error message: 'cannot find attribute `hooks.yamllint.settings.relaxed''
```

I'm sure I'm missing something but from what I've seen in examples I used the modules correctly. Any help is appreciated.

totoroot: @domenkozar

domenkozar:
error: evaluation aborted with the following error message: 'cannot find attribute `hooks.yamllint.settings.relaxed''

modules/hooks.nix
totoroot: I fixed all warnings.

domenkozar:
Ci is failing

PR #404, opened 2024-03-11
Fix typstfmt

NorfairKing:


loicreynier:
Both the binary and the package have indeed been renamed last year.
You should probably also rename the package name in `nix/tools.nix`.

NorfairKing:
@loicreynier done

domenkozar:

PR #405, opened 2024-03-12
typos: allow configPath to be of type string or path + take excludes from `.typos.toml` into account when `pre-commit run typos --all-files` runs

phip1611:
Unfortunately, the `typos` story still isn't solved properly at this point. My concerns decsribed [here](https://github.com/cachix/pre-commit-hooks.nix/pull/392#issuecomment-1893612669) still exist.

## The Problem

_Hint: The problem is also shortly discussed in the commit messages._

Yes, using `pass_filenames = false` should only be done if no other variant is possible (`pass_filenames = true` is fine). My first "fix" merged a few months ago was wrong. And a commit hook should only run on changed files, I agree.

But, people also use `pre-commit run --all-files` to run all checks on the whole repo. **Some people don't even use Git commit hooks at all and just use pre-commit to run all style checks at once.** And that's a perfeclty valid usecase. 

**The problem is** that if you run `$ pre-commit run typos --all-files`, typos behave differently compared to `$ typos .`.

## The Fix

I modified `typos.configPath` in a way to be either of type string or path. The change is additive and not breaking.

To use the new option, you have to do something like this:

```nix
    settings = {
      typos = {
        # configuration = ""  # POSSIBLE
        # configPath = ".typos.toml"; # OLD
        configPath = .typos.toml; # NEW, RECOMMENDED
      };
    };
```

You can verify that my solution works by checking the typos invocation:

`$ strace -s 64 -Tfe trace=execve pre-commit run --all-files typos`

Unintended files are not longer passed to typos. 🎉 

---

Ping @totoroot @domenkozar

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1476,6 +1476,15 @@ in
                 example = "*.nix";
               };
 
+            force-exclude =
+              mkOption {
+                type = types.bool;
+                description = lib.mdDoc "Respect excluded files from config file even for paths passed explicitly";
+                # This must be true as this is the natural behaviour. Similar to when executing `typos`
+                # from the cmdline.
+                default = true;
+              };
+
             format =
               mkOption {
                 type = types.enum [ "silent" "brief" "long" "json" ];
@@ -3375,7 +3384,8 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
                     # Config file always exists (we generate one if not).
                     [ true "--config ${pathToConfigFile}" ]
                     [ diff "--diff" ]
-                    [ (exclude != "") "--exclude ${exclude} --force-exclude" ]
+                    [ (exclude != "") "--exclude ${exclude}" ]
+                    [ (force-exclude) "--force-exclude" ]
                     [ (format != "long") "--format ${format}" ]
                     [ hidden "--hidden" ]
                     [ (locale != "en") "--locale ${locale}" ]
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1436,7 +1436,7 @@ in
             configuration =
               mkOption {
                 type = types.str;
-                description = lib.mdDoc "Multiline-string configuration passed as config file. If set, config set in `typos.settings.configPath` gets ignored.";
+                description = lib.mdDoc "Multiline-string configuration passed as config file. It is recommended to use `configPath` instead for a more natural experience of typos.";
                 default = "";
                 example = ''
                   [files]
@@ -1450,11 +1450,14 @@ in
                 '';
               };
 
+            # It is recommended to use a Nix path here as this way, the excludes
+            # from the config file can be taken into account by pre-commit when
+            # running `$ pre-commit run --all-files`.
             configPath =
               mkOption {
-                type = types.str;
-                description = lib.mdDoc "Path to a custom config file.";
-                default = "";
+                type = types.nullOr (types.either types.str types.path);
+                description = lib.mdDoc "[Path](https://nixos.org/manual/nix/stable/language/values#type-path) to a typos config file (recommended) or a string (deprecated)";
+                default = null;
                 example = ".typos.toml";
               };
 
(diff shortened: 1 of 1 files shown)


phip1611:
FYI ping @blitz

phip1611:
Personally, I'd also like to get rid of `configPath` of type `str`, but then we break the API again..

totoroot:
Sounds good. I'll test it in the coming days.

domenkozar:
Sorry for the mess, we revamed module structure in #397, could you port these changes over?

phip1611:
I'm not sure why GitHub closed this PR.

phip1611:
@domenkozar done

domenkozar:
ping @totoroot for testing

phip1611:
I rebased again and improved my solution. My solution is additive and doesn't break anything; I tested all the existing use cases. Can we merge it, @domenkozar ?

modules/hooks.nix
```diff
@@ -1444,12 +1444,15 @@ in
...
             configPath =
               mkOption {
-                type = types.str;
-                description = lib.mdDoc "Path to a custom config file.";
-                default = "";
-                example = ".typos.toml";
+                type = types.nullOr (types.either types.str types.path);
+                description = lib.mdDoc "Path to a typos config file (recommended) or a string (deprecated).";
+                default = null;
+                example = "./.typos.toml";
```
phip1611, line 1455:
```suggestion
                example = ".typos.toml";
```

modules/hooks.nix
```diff
@@ -3300,6 +3303,45 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
+              builtins.toFile "config.toml"
+                # Concatenate config in config file with section for ignoring words
+                # generated from list of words to ignore
+                ("${hooks.typos.settings.configuration}" +
+                  lib.strings.optionalString (hooks.typos.settings.ignored-words != [ ]) "\n\[default.extend-words\]" +
+                  lib.strings.concatMapStrings (x: "\n${x} = \"${x}\"") hooks.typos.settings.ignored-words
+                )
+          ;
+          # If the config file path is passed as Nix string and not as Nix
+          # Path, we can't read it from Nix, unfortunately.
```
astro, line 3330:
```suggestion
          # If the config file path is passed as string and not as
          # a path, we can't read it from Nix.
```

modules/hooks.nix
```diff
@@ -3300,6 +3303,45 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
+          excludesFromConfig =
+            if builtins.typeOf hooks.typos.settings.configPath == "path" # passed directly or as Path
+            then
+              (
+                let
+                  toml = builtins.fromTOML (builtins.readFile pathToConfigFile);
+                in
+                # The "files.extend-exclude" key comes from
+                  # https://github.com/crate-ci/typos/blob/master/docs/reference.md
```
astro, line 3339: Inconsistent indentation

modules/hooks.nix
```diff
@@ -3300,6 +3303,45 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
+          pathToConfigFile =
+            if hooks.typos.settings.configPath != null
+            # Important: If passed as typeOf == "path", this is in Nix store
+            # which we in fact encourage.
+            then hooks.typos.settings.configPath
+            else
+              builtins.toFile "config.toml"
+                # Concatenate config in config file with section for ignoring words
+                # generated from list of words to ignore
+                ("${hooks.typos.settings.configuration}" +
```
astro, line 3324:
```suggestion
                (hooks.typos.settings.configuration +
```

astro (changes requested):
LGTM apart from a few nits

astro (approved):

phip1611:
Ping @m1-s @domenkozar @totoroot - I hope we can finally get this merged as we need it in multiple of our projects.

domenkozar:
If there's no comments in next few days I'll merge it.

modules/hooks.nix
```diff
@@ -3331,6 +3375,7 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
             in
             "${hooks.typos.package}/bin/typos ${cmdArgs}";
           types = [ "text" ];
+          excludes = excludesFromConfig;
```
roberth, line 3378: `excludes` is expected to be in regex form, whereas `excludesFromConfig` has globs.

modules/hooks.nix
```diff
@@ -1444,11 +1444,14 @@ in
...
+            # It is recommended to use a Nix path here as this way, the excludes
+            # from the config file can be taken into account by pre-commit when
+            # running `$ pre-commit run --all-files`.
             configPath =
               mkOption {
-                type = types.str;
-                description = lib.mdDoc "Path to a custom config file.";
-                default = "";
+                type = types.nullOr (types.either types.str types.path);
+                description = lib.mdDoc "Path to a typos config file (recommended) or a string (deprecated).";
```
roberth, line 1453:
```suggestion
                description = lib.mdDoc "[Path](https://nixos.org/manual/nix/stable/language/values#type-path) to a typos config file (recommended) or a string (deprecated).";
```

roberth (changes requested):
Was mostly interested in the usage of path values, but also found a bug.

modules/hooks.nix
phip1611: From my personal testing I can verify that it works when the `typos.toml` file doesn't use globs.

What can we do do mitigate this issue? This issue really needs to be solved for some of our projects.

modules/hooks.nix
phip1611: thanks!

modules/hooks.nix
roberth: Maybe look for `*`s and `?`s and if they occur, throw an explanation that it's not implemented?
Not fantastic, but leaps better than making users debug or keeping your colleagues/clients waiting.

Ideally, the globs could be translated. https://github.com/hercules-ci/gitignore.nix has inherited an implementation of that from nix-gitignore, but I don't know if it's the same dialect if any. It translates globs into a regex, using a regex... :D ("so now I have three problems"?)

roberth: Are you suggesting to silently drop excludes? I would hate that as a user and flip a table ;)

Likewise!

phip1611:
It's a little unfortunate that the whole topic so complex. Intermediate summary:

- I'm working towards that `$ pre-commit run [--all-files]` behaves similar to `$ typos` with the exact same set of excludes
  - This is the correct thing to do
  - When `typos` is invoked with thousands of files as argument it literally explodes and CPU and mem consumption are insane; system freezes happen (even with `--force-exclude`)
- Typos uses globs, pre-commit uses regex
- If typos "globs without *", it works and it's fine.

The transformation from globs to regex is not trivial.

Observation

1. IMHO this is critical and must be addressed to not annoy and confuse people and to not make them drop typos support in pre-commit hooks at all
2. I do not see a pragmatic way forward here. Such a rather simple thing is already a time sink. I want to satisfy everyones needs. But I can't spend multiple days on this, unfortunately.

This needs to be addressed. The alternative is that we at my Company manually specify excludes locally and that I create a blogpost or so for typos in pre-commit-hooks.nix for other people having the same problem.

ping @roberth - Thoughts?

modules/hooks.nix
phip1611: Agree. Yikes, the whole thing is complicated. I just want `typos` always behave the same, with and without `pre-commit run`.

See https://github.com/cachix/git-hooks.nix/pull/405#issuecomment-2069033631 for a summary of the current discussion.

bmrips:
By now, there is a [`typos.settings.configPath` option](https://github.com/cachix/git-hooks.nix/blob/dcf5072734cb576d2b0c59b2ac44f5050b5eac82/modules/hooks.nix#L1815) which is basically what you were looking for, right?

PR #414, opened 2024-03-22
Added `args` and `exclude_types` to hook type

joedevivo:
[Creating new hooks](https://pre-commit.com/#creating-new-hooks) allows for `args` and `exclude_types`.

In particular, I wanted to configure `shfmt` to not force the `-s` flag. I changed the default configuration for `shfmt` to have `args = [ "-s" ]` to not break anything for anybody, but it allows me to disable it, and add other flags.

### DIFF
--- a/modules/hook.nix
+++ b/modules/hook.nix
@@ -130,6 +130,15 @@ in
       default = [ ];
     };
 
+    exclude_types = mkOption {
+      type = types.listOf types.str;
+      description = lib.mdDoc
+        ''
+          List of file types to exclude. See [Filtering files with types](https://pre-commit.com/#plugins).
+        '';
+      default = [ ];
+    };
+
     pass_filenames = mkOption {
       type = types.bool;
       description = lib.mdDoc ''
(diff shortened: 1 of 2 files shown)


modules/hook.nix
```diff
@@ -130,6 +130,15 @@ in
       default = [ ];
     };
 
+    exclude_types = mkOption {
+      type = types.listOf types.str;
+      description = lib.mdDoc
+        ''
+          List of file types to exclude. See [Filtering files with types](https://pre-commit.com/#plugins).
```
sandydoo, line 137:
```suggestion
          List of file types to exclude. See [Filtering files with types](https://pre-commit.com/#filtering-files-with-types).
```

sandydoo (changes requested):
👍 to both`exclude_types` and `args`.

Regarding `shftm`, `args` is a good fallback/escape-hatch when a hook is missing module options, but we shouldn't make a habit of setting a default of anything other than `[]`. 
Let's maybe expose this as `hooks.shftm.settings.simplify` and set it to `true` by default? Perhaps in a separate PR?

sandydoo:
I've cherry-picked both of these changes out into #480 and #481. Thanks!

PR #415, opened 2024-03-25
Add `check_phase` to pre-commit settings

joedevivo:
The existing behavior of `pre-commit-hooks.nix` is to use the `pre-commit` phase of pre-commit hooks, unless your `default_phases = [ "manual" ]`. On further inspection I found that it has to be only `manual`. If `default_phases = [ "manual" "push" ]`, it will use `commit`. In my particular case, I wanted `default_phases = [ "push" ]`, which then ran `commit` hooks when I ran `nix flake check`.

This PR adds `check_phase` to the configuration. Setting `check_phase = "push"` runs the `pre-push`  phase checks when I run `nix flake check` no matter what my `default_phases` are.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -69,14 +69,14 @@ let
       git config --global user.email "you@example.com"
       git config --global user.name "Your Name"
       git commit -m "init"
-      if [[ ${toString (compare install_stages [ "manual" ])} -eq 0 ]]
-      then
-        echo "Running: $ pre-commit run --hook-stage manual --all-files"
-        ${cfg.package}/bin/pre-commit run --hook-stage manual --all-files
-      else
-        echo "Running: $ pre-commit run --all-files"
-        ${cfg.package}/bin/pre-commit run --all-files
-      fi
+      local check_stage=${toString cfg.check_stage}
+      case $check_stage in
+        commit | merge-commit | push)
+          check_stage="pre-"$check_stage
+          ;;
+      esac
+      echo "Running: $ pre-commit run --hook-stage $check_stage --all-files"
+      ${cfg.package}/bin/pre-commit run --hook-stage $check_stage --all-files
       exitcode=$?
       git --no-pager diff --color
       touch $out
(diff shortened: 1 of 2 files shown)


sandydoo:
>  On further inspection I found that it has to be only manual. If default_phases = [ "manual" "push" ], it will use commit.  In my particular case, I wanted default_phases = [ "push" ], which then ran commit hooks when I ran nix flake check.

@joedevivo, if I understand correctly, the issue you were having was that, after setting `default_phases = ["manual" "push"]`, the `pre-commit` hooks still ran before each commit. Is that correct?
This sounds like a bug, which might be fixed #423.

As far as the feature itself goes, the `manual` stage/phase we had before feels like the appropriate choice for `nix flake check`. Even if we expose this option (and I'm not entirely sure what the use-case for that would be), the default should remain `manual`.

m1-s:
Added the mentioned test cases to the new test suite https://github.com/cachix/git-hooks.nix/pull/423/files#diff-6c8d597050bd88f356507ec90b216be2fcd69fce6e488c62060f12bae96deddeR37-R51

Indeed my PR appears to fix the mentioned issue already. Let me know if I misunderstood anything.

modules/pre-commit.nix
```diff
@@ -263,6 +263,16 @@ in
           default = [ "commit" ];
         };
 
+      check_stage =
+        mkOption {
+          type = types.str;
```
ThunderKey, line 268: This could be the enum part of `supportedHooksLib.supportedHooksType`

modules/pre-commit.nix
```diff
@@ -263,6 +263,16 @@ in
...
+          description = lib.mdDoc
+            ''
+              This is pre-commit stage to be used in a nix `checkPhase`. `commit` maps to `pre-commit`, which is `pre-commit`'s default.
+            '';
+          default = "commit";
```
ThunderKey, line 273: We could set this to the first of `default_stages` and fallback to `pre-commit`

sandydoo:
Superseded by #689.

PR #416, opened 2024-03-26
Add Poetry hooks

loicreynier:
Add Poetry hooks to check the configuration file and the state of the lock files. 

Closure size is `3.7M` according to` du -sh $(nix-build -A typos)`.


totoroot (approved):
This would be incredibly useful. Thanks @loicreynier!

PR #419, opened 2024-03-27
feat: Change `terraform-format` to support both Terraform and OpenTofu packages

### LINKED ISSUE #418: Change `terraform-format` to support both Terraform and OpenTofu packages
Right now, the `terraform-fmt` hook uses OpenTofu under the hood, which is a little misleading given the hook name. On the other hand, the use of OpenTofu over Terraform may be something to leave to the end user to decide. ~~In that regard, I suggest creating two separate hooks, one that uses Terraform and another that uses OpenTofu.~~

sestrella:
Fixes #418


.github/workflows/ci.yaml
```diff
@@ -15,7 +15,9 @@ jobs:
       with:
         name: pre-commit-hooks
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-    - run: nix-build --keep-going
+    - run: nix-build --keep-going --impure
+      env:
+        NIXPKGS_ALLOW_UNFREE: 1
```
sestrella, line 20: I'm not a huge fan of turning on this flag, but the Terraform version available in the current `nixpkgs` requires it. We could also try to pull an older version of Terraform like `1.5.7` before HashiCorp made the license change, but that would require setting up a new input like [nixpkgs-terraform](https://github.com/stackbuilders/nixpkgs-terraform).

modules/hooks.nix
```diff
@@ -2848,9 +2856,9 @@ in
       terraform-format =
         {
           name = "terraform-format";
-          description = "Format terraform (`.tf`) files.";
-          package = tools.terraform-fmt;
-          entry = "${hooks.terraform-format.package}/bin/terraform-fmt";
+          description = "Format Terraform (`.tf`) files.";
+          package = tools.terraform;
+          entry = "${hooks.terraform-format.package}/bin/terraform fmt -check -diff";
```
sestrella, line 2861: I realized that the `terraform-fmt` script was not necessary. I tried calling directly  `terraform fmt -check -diff` from the hook `entry` and it worked fine. I created a test [repository](https://github.com/sestrella/test-pre-commit-hooks.nix/blob/main/devenv.yaml) to validate this. Here is the outcome of the `devenv test` command running on the CI:

image

.github/workflows/ci.yaml
totoroot: That is why I opted for `opentofu` instead of `terraform` in the `aws-vault` hook. I feel like most users would prefer using the Open Source fork, but I get what you are saying that the naming might be confusing.

I don't really use the hook myself and was just making changes to fix the CI build of `aws-vault` in https://github.com/cachix/devenv/pull/994, so I'm impartial to using one or the other.

However, I don't think that we should purposely use a pre-licence-change version of terraform in an separate input for the hook.

If users wish to use an old version that was released under a true Open Source, then I'd say they should just use a custom hook.

sestrella: Fair point; another option that I considered was to write a script that checks if the `tofu` executable is available and falls back to `terraform` if it isn't, something like this:

```nix
if [ -f ${hooks.terraform-format.package}/bin/tofu ]; then
  ${hooks.terraform-format.package}/bin/tofu fmt -check -diff "$@"
else
  ${hooks.terraform-format.package}/bin/terraform fmt -check -diff "$@"
fi
```

This way, we can keep the current implementation while making it easier for users to switch to Terraform if they want to. Part of the reason I submitted this PR in the first place is that it feels strange to use OpenToFu to look for formatting errors when using `languages.terraform.enable` on `devenv`; ideally, this hook should use the same version of OpenTofu or Terraform as `languages.opentofu.package` or `languages.terraform.package` respectively. The changes made in this PR would make it easier to mimic some of the behaviour found in the [languages.rust](https://github.com/cachix/devenv/blob/main/src/modules/languages/rust.nix#L92-L94) module within `devenv`, where the module passes the appropriate tooling to the `pre-commit` hooks to avoid installing multiple versions of the same package.

modules/hooks.nix
```diff
@@ -3225,9 +3225,26 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
-          package = tools.terraform-fmt;
-          entry = "${hooks.terraform-format.package}/bin/terraform-fmt";
+          description = "Format Terraform (`.tf`) files.";
+          package = tools.opentofu;
+          entry =
+            let
+              terraform-fmt = pkgs.writeScriptBin "terraform-fmt" ''
+                #!/usr/bin/env bash
+
+                opentofu_or_terraform() {
```
sestrella, line 3235: @totoroot Based on our discussion, I made a few changes to the existing `terraform-format`. For the time being, I feel it would be easier to alter the existing hook to work with both OpenTofu and Terraform. End users can choose whether to use OpenTofu (the default) or alter the `package` to Terraform, as shown in this example:

https://github.com/sestrella/test-pre-commit-hooks.nix/blob/main/devenv.nix

I'm planning to update the PR title and the issue accordingly.

sestrella:
@domenkozar I just resolved the conflicts on this branch and updated the repo I was using for testing; so far, it appears to be working properly:

```
--- devenv:git-hooks:run failed with error: Task exited with status: exit status: 1
--- devenv:git-hooks:run stdout:
0000.02: terraform-format.........................................................Failed
0000.02: - hook id: terraform-format
0000.02: - exit code: 3
0000.02: 
0000.02: main.tf
0000.02: --- old/main.tf
0000.02: +++ new/main.tf
0000.02: @@ -1,4 +1,4 @@
0000.02:  locals {
0000.02: -  foo     = 1
0000.02: -  bar   = 2
0000.02: +  foo = 1
0000.02: +  bar = 2
0000.02:  }
0000.02: test.tf
0000.02: --- old/test.tf
0000.02: +++ new/test.tf
0000.02: @@ -1,4 +1,4 @@
0000.02:  locals {
0000.02: -  zoo = 1
0000.02: +  zoo    = 1
0000.02:    asdasd = 2
0000.02:  }
0000.02: 
--- devenv:git-hooks:run stderr:
```

Would you mind taking a look, please?

https://github.com/sestrella/test-pre-commit-hooks.nix/actions/runs/12173105536/job/33952913600

sestrella:
I refactored the `terraform-format` hook to fully remove the custom script, similar to #473. Furthermore, the changes on this PR allow to use either `opentofu` or `terraform` to check for formatting errors in Terraform files. Here's an example of the hook in use:

```
0000.02: terraform-format.........................................................Failed
0000.02: - hook id: terraform-format
0000.02: - exit code: 3
0000.02: 
0000.02: foo.tf
0000.02: --- old/foo.tf
0000.02: +++ new/foo.tf
0000.02: @@ -1,4 +1,4 @@
0000.02:  locals {
0000.02: -  asdf = 1
0000.02: +  asdf    = 1
0000.02:    asddddd = 2
0000.02:  }
0000.02: test.tf
0000.02: --- old/test.tf
0000.02: +++ new/test.tf
0000.02: @@ -1,4 +1,4 @@
0000.02:  locals {
0000.02: -  zoo = 1
0000.02: +  zoo    = 1
0000.02:    asdasd = 2
0000.02:  }
0000.02: 
```

Source: https://github.com/sestrella/test-pre-commit-hooks.nix/actions/runs/12616401258/job/35157601036#step:6:1073

PR #421, opened 2024-03-29
Fix `nix flake check` on macOS

9999years:
This fixes `nix flake check` on macOS.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,9 @@
         };
 
         checks = lib.filterAttrs (k: v: v != null)
-          (exposed.checks // exposed-stable.checks);
+          (exposed.checks
+          // (lib.mapAttrs' (name: value: lib.nameValuePair "stable-${name}" value)
+            exposed-stable.checks));
 
         lib = { inherit (exposed) run; };
       }
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -7,13 +7,14 @@ let
   overlay =
     self: pkgs:
     let
+      inherit (pkgs) lib;
       tools = import ./call-tools.nix pkgs;
       run = pkgs.callPackage ./run.nix { inherit pkgs tools isFlakes gitignore-nix-src; };
     in
     {
       inherit tools run;
       # Flake style attributes
-      packages = tools // {
+      packages = (lib.filterAttrs (_name: value: value != null) tools) // {
         inherit (pkgs) pre-commit;
       };
       checks = self.packages // {
@@ -27,7 +28,7 @@ let
         };
         all-tools-eval =
           let
-            config = pkgs.lib.evalModules {
+            config = lib.evalModules {
               modules = [
                 ../modules/all-modules.nix
                 {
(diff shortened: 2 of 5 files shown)


flake.nix
```diff
@@ -48,7 +48,9 @@
         };
 
         checks = lib.filterAttrs (k: v: v != null)
-          (exposed.checks // exposed-stable.checks);
+          (exposed.checks
+          // (lib.mapAttrs' (name: value: lib.nameValuePair "stable-${name}" value)
+            exposed-stable.checks));
```
9999years, line 53: These two sets had the same names, so only the checks from the second set (`exposed-stable`) were being built. This led to a confusing situation where `nix flake check` would fail but building the corresponding package with `nix build .#opam` would succeed.

nix/default.nix
```diff
@@ -7,13 +7,14 @@ let
...
     let
+      inherit (pkgs) lib;
       tools = import ./call-tools.nix pkgs;
       run = pkgs.callPackage ./run.nix { inherit pkgs tools isFlakes gitignore-nix-src; };
     in
     {
       inherit tools run;
       # Flake style attributes
-      packages = tools // {
+      packages = (lib.filterAttrs (_name: value: value != null) tools) // {
```
9999years, line 17: `nix flake check` complains if non-derivation values are present in `packages`.

nix/default.nix
```diff
@@ -37,7 +38,10 @@ let
               specialArgs = { inherit pkgs; };
             };
             allHooks = config.config.hooks;
-            allEntryPoints = pkgs.lib.mapAttrsToList (_: v: v.entry) allHooks;
+            allEntryPoints = lib.pipe allHooks [
+              (lib.filterAttrs (_name: value: value.package != null))
```
9999years, line 42: Can't coerce `null` to a string.

nix/headache/default.nix
```diff
@@ -1,9 +1,16 @@
...
 
 ## NOTE: `headache` moved from `ocamlPackages.headache` to top-level on 8 June
 ## 2023. Once this gets into a NixOS release, the following code will be
 ## useless.
-let the-headache = if headache != null then headache else ocamlPackages.headache; in
+let
+  the-headache =
+    (if headache != null then headache else ocamlPackages.headache).overrideAttrs (drv: {
+      nativeBuildInputs = (drv.nativeBuildInputs or [ ]) ++ lib.optionals stdenv.isDarwin [
+        darwin.sigtool
```
9999years, line 10: Missing `codesign` on nixpkgs-stable.

nix/tools.nix
```diff
@@ -183,6 +174,9 @@ in
   chktex = tex;
   commitizen = commitizen.overrideAttrs (_: _: { doCheck = false; });
   bats = if bats ? withLibraries then (bats.withLibraries (p: [ p.bats-support p.bats-assert p.bats-file ])) else bats;
+  git-annex = if stdenv.isDarwin then null else git-annex;
```
9999years, line 177: Setting these to `null` is kinda sketchy (`meta.broken` would be better) but it's what the other derivations in this file do so I left it as-is.

PR #422, opened 2024-04-01
Option for `hookModule`

roberth:
This removes a bit of duplication, and allows `hookModule` to be monkey patched.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1,14 +1,8 @@
-{ config, lib, pkgs, ... }:
+{ config, lib, pkgs, hookModule, ... }:
 let
   inherit (config) hooks tools settings;
   cfg = config;
   inherit (lib) flatten mapAttrs mapAttrsToList mkDefault mkOption mkRenamedOptionModule types;
-
-  hookModule =
-    [
-      ({ ... }: { _module.args.default_stages = cfg.default_stages; })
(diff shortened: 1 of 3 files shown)


modules/hooks.nix
```diff
@@ -44,6 +38,20 @@ in
...
+      Base module that must be loaded into each hook.
+    '';
+  };
+
+  config.hookModule = {
+    imports = [ ./hook.nix ];
+    config._module.args.default_stages = cfg.default_stages;
+  };
+  config._module.args.hookModule = config.hookModule;
+
```
roberth, line 54: This chunk could be moved out to a separate file if you like.

roberth:
> `_module.args` are way more powerful than I initially thought.


You can't use it for imports in the same fixpoint (here: the root, or initial `evalModules`), but you can refer to it in submodules. Those are independent `evalModules` calls performed by the submodule types.

sandydoo:
@azahi, fixed

SebTM:
Thanks for the fast fix 🥇 Works fine again on `8ae5bb63f9bcd0472b67cf4ada414ae37f1f2bcc` (latest master) with both options:

```
inherit (self.checks.${system}.pre-commit-check) shellHook;
buildInputs = self.checks.${system}.pre-commit-check.enabledPackages;
```

PR #423, opened 2024-04-01
add hook installation test

m1-s:
This PR adds a test that checks that the correct hooks are installed in the hooks directory.

While writing the tests I already discovered and fixed a bug (I think) in this line:
```nix
install_stages = lib.unique (cfg.default_stages ++ (builtins.concatLists (lib.mapAttrsToList (_: h: h.stages) enabledHooks)));
```

This line creates the default stage `pre-commit` even when `stages` is manually set to be something else. See the diff for the fix.

This work was started on the #Thaigersprint.

### DIFF
--- a/nix/installation-test.nix
+++ b/nix/installation-test.nix
@@ -1,0 +1,99 @@
+# Test that checks whether the correct hooks are created in the hooks folder.
+{ git, perl, coreutils, runCommand, run, lib, mktemp }:
+let
+  tests = {
+    basic-test = {
+      expectedHooks = [ "pre-commit" ];
+      conf.hooks.shellcheck.enable = true;
+    };
+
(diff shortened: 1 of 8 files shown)


nix/installation-test.nix
```diff
@@ -0,0 +1,74 @@
...
+      conf = {
+        default_stages = [ "commit-msg" ];
+        hooks.nixpkgs-fmt.enable = true;
+      };
+    };
+  };
+
+  testsInShell = lib.mapAttrsToList
+    (name: test:
+      let runDerivation = run ({ src = ../.; } // test.conf);
```
m1-s, line 40: Use an empty folder for src.

m1-s:
fixed mac-os pipeline. pls restart workflow.

PR #432, opened 2024-04-13
feat: create "biome" hook, preserve "rome" hook as alias

### LINKED ISSUE #428: `biome` is being used instead of `rome`
While navigating the repository to implement https://github.com/cachix/git-hooks.nix/issues/425, I noticed that the `rome` hook is already configured to use `biome` instead of the actual `rome` tool.

This changed was performed in https://github.com/cachix/git-hooks.nix/pull/391, and `biome` is still referenced and used as `rome` throughout the documentation and code.

I propose to rename the `rome` hook to `biome` and create an alias to `rome` for backwards compatibility.

oxcabe:
This PR does the following:
* Renames the "rome" hook to "biome", correcting any references to the rome tool.
* Sets the "rome" hook as an alias to the "biome" hook as a backwards compatibility guard.
* Adds relevant documentation to the README.

Fixes #428.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -73,7 +73,7 @@ in
   };
 
   # PLEASE keep this sorted alphabetically.
-  options.hooks =
+  options.hooks = rec
     {
       alejandra = mkOption {
         description = lib.mdDoc "alejandra hook";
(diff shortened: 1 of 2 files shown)


sandydoo (changes requested):
@oxcabe, could you please remove all the code formatting changes? The diff should be readable and contain just the changes you're proposing. Thanks!

oxcabe:
Sure thing, I'm not even sure when the files got formatted :sweat_smile:

sandydoo (changes requested):
I think the better approach here would be to use `mkRenamedOptionModule` to rename `rome -> biome`. You get the alias and people will be prompted to switch.

If you feel like an alias (with no warning) is more appropriate for now, then you can use `mkAliasOptionModule` instead. 

There's no need to have both `rome` and `biome` in `config.hooks`.

oxcabe:
I know it's failing as of my last commit. Currently trying to figure out by myself what am I doing wrong and why 👍🏼

oxcabe:
> I think the better approach here would be to use mkRenamedOptionModule to rename rome -> biome. You get the alias and people will be prompted to switch.

Implemented this approach. For some reason, the deprecation warning is not appearing when enabling the hook:

```
rome.enable = true;
```

> There's no need to have both rome and biome in config.hooks.

I can't seem to make it work without having `config.hooks` entries for both tools.

modules/hooks.nix
```diff
@@ -1638,7 +1640,7 @@ in
     };
```
sandydoo, line 1641: Ah, okay, we're hitting the limitations of the freetype+submodule system. We should manually add a warning to help people switch over to biome. Something like this:

```suggestion

  config.warnings =
    lib.optional cfg.hooks.rome.enable ''
      The hook `hooks.rome` has been renamed to `hooks.biome`.
    '';

```

sandydoo (changes requested):

oxcabe:
From the failing build:

```
pre-commit-run> - hook id: nixpkgs-fmt
pre-commit-run> - files were modified by this hook
```

modules/hooks.nix
oxcabe: I like this approach :+1:

oxcabe:
Ready!

sandydoo (approved):

PR #433, opened 2024-04-16
fix(alejandra): one `--exclude` flag per file

mrcjkb:
`alejandra --exclude file1 file2 file3` doesn't work.

It should be `alejandra --exclude file1 --exclude file2 --exclude file3 `

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1659,7 +1659,7 @@ in
               cmdArgs =
                 mkCmdArgs (with hooks.alejandra.settings; [
                   [ check "--check" ]
-                  [ (exclude != [ ]) "--exclude ${lib.escapeShellArgs (lib.unique exclude)}" ]
+                  [ (exclude != [ ]) "${lib.escapeShellArgs (map (file: "--exclude '${file}'") (lib.unique exclude))}" ]
                   [ (verbosity == "quiet") "-q" ]
                   [ (verbosity == "silent") "-qq" ]
                   [ (threads != null) "--threads ${toString threads}" ]


sandydoo (approved):

modules/hooks.nix
```diff
@@ -1659,7 +1659,7 @@ in
               cmdArgs =
                 mkCmdArgs (with hooks.alejandra.settings; [
                   [ check "--check" ]
-                  [ (exclude != [ ]) "--exclude ${lib.escapeShellArgs (lib.unique exclude)}" ]
+                  [ (exclude != [ ]) "${lib.escapeShellArgs (map (file: "--exclude '${file}'") (lib.unique exclude))}" ]
```
roberth, line 1662: `exclude` has a specific meaning in `pre-commit`, which is probably not this.

Wouldn't it be better to let pre-commit pass the list of changed files anyway?

mrcjkb: Not sure I understand.
Afaik, pre-commit already passes in a list of changed files.
My use case for `--exclude` with alejandra is to prevent it from formatting nix files that were generated (e.g. with a cabal2nix hook).

mrcjkb: I suppose for the use case of generating nix files with cabal2nix, it would make sense to use

> git-hooks.nix: `excludes`, plural, regex

I'll give it a go in a few days and report back here.

mrcjkb: I've just tested the git-hooks excludes, and it works fine for my use case.
So I would agree that alejandra.settings.excludes is redundant.

PR #435, opened 2024-04-18
Add Flake Checker

loicreynier:
Add Flake Checker from [Determinate Systems](https://github.com/DeterminateSystems/flake-checker) to run health check on the Nix flake.

Closure size is `11M` according to `du -sh $(nix-build -A flake-checker)`.


sandydoo (changes requested):
Otherwise LGTM

PR #442, opened 2024-05-07
Add ltex-ls hook

michaelvanstraten:


michaelvanstraten:
Draft because `ltex-ls` [currently throws an exception](https://github.com/valentjn/ltex-ls/issues/303) when errors span multiple lines.

Issue #445, opened 2024-05-15
Can the number of inputs/dependencies be lowered for the Flake?


toastal:
I have included a different project into my flake that had this project as a input & now my flake lock has exploded with new things I need to audit. Are all of these inputs actually required?

sandydoo:
Could you please elaborate on this? What is your definition of "exploded" and which inputs do you consider redundant?

We're aware that the project ships with two versions of nixpkgs. We'd like to remove them at some point. Other than that, there's only 5 inputs...

sandydoo:
Sure, we can replace `flake-utils` with the common `forAllSystems` function. I can merge that in, PRs welcome.

PR #446, opened 2024-05-15
Drop `flake-utils` dependency

toastal:
I wanted to run `nix flake checks` but it was going to download packages for hours on my 3G-tethered connection … where 2 `nixpkgs` was enough data usage & all these dependencies really start to add up (hence opening the initial issue).

Partially addresses #445

Wasn’t entirely sure how to go about the `README`. After some thought today, I’m not convinced its good to recommend ‘looping’ in examples as folks might be accidentally supporting platforms they or their tools can’t & `nix flake init` just hardcodes the current system to keep everything simple. I went this route with `${system}` as a variable as the block can still be copied into someone’s `(system: …)` ‘loop’, but not married to the idea.

Additionally, removed the unused C compiler from the devShell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,25 +29,30 @@ Given the following `flake.nix` example:
   description = "An example project.";
 
   inputs.pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
 
-  outputs = { self, nixpkgs, pre-commit-hooks, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
-      {
-        checks = {
-          pre-commit-check = pre-commit-hooks.lib.${system}.run {
-            src = ./.;
-            hooks = {
-              nixpkgs-fmt.enable = true;
-            };
+  outputs = { self, nixpkgs, ... }@inputs:
+    let
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+
+      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+    in
+    {
+      checks = forAllSystems (system: {
+        pre-commit-check = inputs.pre-commit-hooks.lib.${system}.run {
+          src = ./.;
+          hooks = {
+            nixpkgs-fmt.enable = true;
           };
         };
-        devShell = nixpkgs.legacyPackages.${system}.mkShell {
+      });
+
+      devShells = forAllSystems (system: {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,25 +29,28 @@ Given the following `flake.nix` example:
...
         };
-      }
-    );
+      };
+    };
+    devShell = nixpkgs.legacyPackages.${system}.mkShell {
+      inherit (self.checks.${system}.pre-commit-check) shellHook;
+      buildInputs = self.checks.${system}.pre-commit-check.enabledPackages;
+    };
+  }
```
sandydoo, line 53: Lets revert this.

1. This isn't a valid flake.
2. People copy-paste these snippets all the time. It should at least evaluate on all 4 default systems out of the box. We get enough support requests as it is!
3. We're not the place to teach people how to use flakes. The main point here is to show a reasonable flake config to get people started with git hooks.

sandydoo (changes requested):

flake.nix
```diff
@@ -22,6 +21,12 @@
         "x86_64-darwin"
         "x86_64-linux"
       ];
+      depsFor = lib.genAttrs defaultSystems (system: {
+        pkgs = import nixpkgs { inherit system; };
```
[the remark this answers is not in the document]
sandydoo: We can reuse the same evaluation of the nixpkgs set between flakes. If every flake does its own import and re-evaluates nixpkgs, we quickly create a huge performance issue.

README.md
sandydoo: > My concern is folks copying the snippet then all of their downstream users will have a probably-unnecessary dependency to audit from copying code from the repository.

A laudable cause, but it feels as if we're trying to patch the design deficiencies of flakes by making everyone else miserable.

> Are you sure this is preferred to setting the example without `flake-utils` with a `nixpkgs.lib.genAttrs nixpkgs.lib.systems.flakeExposed`? 

Honestly, I don't think it really matters in the grand scheme of things. Again, this is an example meant to showcase the pre-commit/git-hooks bits.

We could do a `forAllSystems = nixpkgs.lib.genAttrs systems`, where `systems = [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];`? This is generally what we do across cachix repos. `flakeExposed` seems like way too much, no?
```
nix-repl> lib.systems.flakeExposed
[ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "armv6l-linux" "armv7l-linux" "i686-linux" "mipsel-linux" "aarch64-darwin" "armv5tel-linux" "powerpc64le-linux" "riscv64-linux" ]
```

Issue #448, opened 2024-05-19
`enabledPackages` does not include `hpack` if the hpack hook is enabled.


NorfairKing:
This is because the hpack hook uses a `hpack-dir` executable.

Can we somehow `hpack` in the `hpack-dir` package, for example?

PR #449, opened 2024-05-19
Include hpack in the hpack dir output

NorfairKing:
Fix for #448

### DIFF
--- a/nix/hpack-dir/default.nix
+++ b/nix/hpack-dir/default.nix
@@ -1,6 +1,6 @@
 { writeScriptBin, hpack }:
 
-writeScriptBin "hpack-dir" ''#!/usr/bin/env bash
+(writeScriptBin "hpack-dir" ''#!/usr/bin/env bash
   set -e
   ##  ^^
   ## The `-e` flag changes the behaviour of Shell so that the first top-level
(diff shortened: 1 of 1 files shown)


sandydoo (approved):
This'll work 👍 I suppose `symlinkJoin` is also an option.

We now have [`extraPackages`](https://github.com/cachix/git-hooks.nix/blob/fa606cccd7b0ccebe2880051208e4a0f61bfc8c1/modules/hooks.nix#L1352-L1355) to accomplish this. But...since we don't have an easy way to override `hpack`, I think it would be deceiving to use it until we first untangle `tools`.

NorfairKing:
@sandydoo would you prefer the `extraPackages` approach? I guess I do care about overriding but not that much..

sandydoo:
> @sandydoo would you prefer the `extraPackages` approach? I guess I do care about overriding but not that much..

@NorfairKing, sure

NorfairKing:
@sandydoo I'm thinking about it again and making overriding `hpack` possible is actually pretty important in hindsight. Otherwise there's just no way to do it.

I'm open to nicer approaches though.

NorfairKing:
@sandydoo Thanks! I'd be happy to help maintain this hook in case we run into trouble because of this.

PR #454, opened 2024-06-06
feat: add typstyle

leana8959:
Hello,

I have added `typstyle`, a formatter for `typst`.

Though, there's a little issue: it's not in the version of nixpkgs pinned by this project, and it won't work. What would be a good solution ?

Thanks and have a nice day :)

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -3418,7 +3418,11 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
         name = "typstyle";
         description = "Beautiful and reliable typst code formatter";
         packages = tools.typstyle;
-        entry = "${hooks.typstyle.package}/bin/typstyle";
+        entry =
+          lib.throwIf
+            (hooks.typstyle.package == null)
+            "The version of nixpkgs used by pre-commit-hooks.nix must contain typstyle"
+            "${hooks.typstyle.package}/bin/typstyle";
         files = "\\.typ$";
       };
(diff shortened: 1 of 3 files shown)


leana8959:
_Psst, The CI failed because `typstyle` is not in the `nixpkgs` pinned in the flake._

nix/tools.nix
```diff
@@ -71,6 +71,7 @@
 , treefmt
 , typos
 , typstfmt
+, typstyle
```
sandydoo, line 74:
```suggestion
, typstyle ? null
```

modules/hooks.nix
```diff
@@ -3414,6 +3414,13 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
         entry = "${hooks.typstfmt.package}/bin/typstfmt";
         files = "\\.typ$";
       };
+      typstyle = {
+        name = "typstyle";
+        description = "Beautiful and reliable typst code formatter";
+        packages = tools.typstyle;
+        entry = "${hooks.typstyle.package}/bin/typstyle";
```
sandydoo, line 3421: You can check whether `tools.typstyle` is `null` and print a nicer error message if it is. Search for `lib.throwIf` for examples.

sandydoo (changes requested):
I'll bump nixpkgs so that this works out of the box.

leana8959:
The requested changes were added :)

sandydoo:
Thanks, @leana8959!

PR #469, opened 2024-06-27
docs: clarify terraform-format

joscha:
Adds a hint to the documentation that the built-in `terraform-format` hook uses OpenTofu under the hood.


joscha:
thank you @sandydoo

PR #473, opened 2024-07-15
Remove `terraform-fmt` wrapper

lbodor:
Since the time `terraform-fmt` wrapper was added to `git-hooks.nix` 5 years ago, both `terraform` and `tofu` now support formatting multiple files.

https://github.com/hashicorp/terraform/pull/31286
https://github.com/hashicorp/terraform/pull/28191


lbodor:
I just noticed https://github.com/cachix/git-hooks.nix/pull/419, which also makes the change to format individual files, rather than entire directories, like the existing `terraform-fmt` wrapper.

domenkozar:
Should we rather use #419?

lbodor:
> Should we rather use #419?

Sure, that would be fine.

sandydoo:
Merged #419, which is practically identical to this.

PR #480, opened 2024-08-06
feat: add `args` and `exclude_types` to hooks

sandydoo:
Adds `args` —  a list of additional parameters to pass to the hook.
Adds `exclude_types` —  list of file types to exclude.

Cherry-picked from #414.


sandydoo:
`args` is quite helpful in situations like https://github.com/cachix/git-hooks.nix/issues/477. You can just add the arguments you need without waiting for us to do it.

PR #488, opened 2024-08-23
docs: rename `pre-commit-hooks.nix` instances to `git-hooks.nix`

trueNAHO:
Rename all `pre-commit-hooks.nix` instances (except in `https://flake.parts/options/pre-commit-hooks-nix`) to `git-hooks.nix` using the following command:

```bash
fd --type f --exec-batch sed --in-place 's/pre-commit-hooks\.nix/git-hooks.nix/g'
```

This change leaves all remaining `pre-commit-hooks` instances untouched.


domenkozar:
Thanks!

PR #493, opened 2024-08-29
nixfmt: deprecate `nixfmt` and redirect to `nixfmt-classic` or `nixfmt-rfc-style`

sandydoo:
Final bit of migration for #492.

```nix
{ 
  # The existing hook is deprecated
  hooks.nixfmt.enable = true;
  
  # and should be changed to
  hooks.nixfmt-classic.enable = true;
  # or
  hooks.nixfmt-rfc-style.enable = true;
}
```

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -846,7 +846,22 @@ in
         };
       };
       nixfmt = mkOption {
-        description = "nixfmt hook";
+        description = "Deprecated nixfmt hook";
+        visible = false;
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.settings = {
+            width =
+              mkOption {
+                type = types.nullOr types.int;
+                description = "Line width.";
+                default = null;
+              };
+          };
+        };
+      };
+      nixfmt-classic = mkOption {
+        description = "nixfmt (classic) hook";
         type = types.submodule {
           imports = [ hookModule ];
           options.settings = {
@@ -1721,6 +1736,11 @@ in
   config.warnings =
     lib.optional cfg.hooks.rome.enable ''
       The hook `hooks.rome` has been renamed to `hooks.biome`.
+    ''
+    ++ lib.optional cfg.hooks.nixfmt.enable ''
+      The hook `hooks.nixfmt` has been renamed to `hooks.nixfmt-classic`.
+
+      The new RFC 166-style nixfmt is available as `hooks.nixfmt-rfc-style`.
     '';
 
   # PLEASE keep this sorted alphabetically.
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -2874,12 +2894,15 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
+      nixfmt = nixfmt-classic;
+      nixfmt-classic =
         {
-          name = "nixfmt";
-          description = "Nix code prettifier.";
+          name = "nixfmt-classic";
+          description = "Nix code prettifier (classic).";
           package = tools.nixfmt-classic;
-          entry = "${hooks.nixfmt.package}/bin/nixfmt ${lib.optionalString (hooks.nixfmt.settings.width != null) "--width=${toString hooks.nixfmt.settings.width}"}";
+          entry = "${hooks.nixfmt-classic.package}/bin/nixfmt ${lib.optionalString (hooks.nixfmt-classic.settings.width != null) "--width=${toString hooks.nixfmt-classic.settings.width}"}";
```
MattSturgeon, line 2905: This line lead to a [regression](https://github.com/nix-community/nixvim/pull/2188#issuecomment-2336471966) in nixvim, where we were using `nixfmt.enable`, but preemptively setting `nixfmt.package = pkgs.nixfmt-rfc-style`.

Because the `nixfmt` hook is now a copy of `nixfmt-classic` (but not a proper two-way alias), this line is ignoring any config under `nixfmt.*` and only considering `nixfmt-classic.*`.

I think this could be solved by using `mkAliasOptionModule` or `mkRenamedOptionModule` instead of doing a simple copy on line 2899.

E.g:

```nix
imports = [
  (lib.mkAliasOptionModule [ "hooks" "nixfmt" ] [ "hooks" "nixfmt-classic" ])
];
```

This way any configuration defined under the `hooks.nixfmt` option will be forwarded to the `nixfmt-classic` option (and vice-versa).

sandydoo: Thanks for the feedback, and sorry for causing trouble!

Unfortunately, we can't use `mkAliasOptionModule` or `mkRenamedOptionModule` because `hooks` is a freeform set of submodule. See https://github.com/NixOS/nixpkgs/issues/96006.

We'll need to add some machinery to handle this properly.

MattSturgeon: Thanks for the added context!

I believe that issue only relates to the warnings produced by `mkRenamedOptionModule` not being able to "escape" the submodule. If I understand correctly, `mkAliasOptionModule` would still work as intended.

You could of course do things a bit more manually, for example with `mkAliasAndWrapDefinitions`, although if `hooks` is a submodule I think that'd be easier to do _within_ the submodule:

```nix
{ lib, config, options, ... }:
{
  config = lib.mkIf config.nixfmt.enable {
    nixfmt-classic = lib.mkAliasAndWrapDefinitions lib.id options.nixfmt;

    # Or if that is recursive:
    nixfmt-classic = builtins.removeAttrs (lib.mkAliasAndWrapDefinitions lib.id options.nixfmt) [ "enable" ];
  }
}
```

But I think `mkAliasOptionModule` should be fine :thinking:

sandydoo: You're absolutely right. I just realised we've been trying to apply the renames to the top-level options, which is why the `toRename` machinery would always fail to evaluate. Once applied to the submodule, it works as excepted.

PR #495, opened 2024-09-20
Add nix fmt hook

shikanime:


modules/hooks.nix
```diff
@@ -49,7 +49,7 @@ in
     ++ [ (mkRenamedOptionModule [ "settings" "rome" ] [ "hooks" "biome" "settings" ]) ]
     # Rename the remaining `settings.<name>` to `hooks.<name>.settings`
     ++ map (name: mkRenamedOptionModule [ "settings" name ] [ "hooks" name "settings" ])
-      [ "ansible-lint" "autoflake" "biome" "clippy" "cmake-format" "credo" "deadnix" "denofmt" "denolint" "dune-fmt" "eslint" "flake8" "headache" "hlint" "hpack" "isort" "latexindent" "lychee" "mkdocs-linkcheck" "mypy" "nixfmt" "ormolu" "php-cs-fixer" "phpcbf" "phpcs" "phpstan" "prettier" "psalm" "pylint" "pyright" "pyupgrade" "revive" "statix" ];
+      [ "ansible-lint" "autoflake" "biome" "clippy" "cmake-format" "credo" "deadnix" "denofmt" "denolint" "dune-fmt" "eslint" "flake8" "headache" "hlint" "hpack" "isort" "latexindent" "lychee" "mkdocs-linkcheck" "mypy" "nix" "nixfmt" "ormolu" "php-cs-fixer" "phpcbf" "phpcs" "phpstan" "prettier" "psalm" "pylint" "pyright" "pyupgrade" "revive" "statix" ];
```
sandydoo, line 52: What option are we renaming here? I don't think we need this.

modules/hooks.nix
```diff
@@ -2894,6 +2894,13 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
             builtins.toString script;
           files = "\\.nix$";
         };
+      nix-fmt =
+        {
+          name = "nix-fmt";
+          description = "Nix formatter.";
```
sandydoo, line 2900: Lets come up with a better description here, otherwise this looks like another formatter, instead of something that calls a formatter from the flake.

Maybe something along the lines of "Format Nix files using the formatter specified in your flake."?

sandydoo (changes requested):

sandydoo (approved):

PR #496, opened 2024-09-26
fix: add configFile option

### LINKED ISSUE #345: [question] Can I use this to override a hook while leaving `pre-commit-hooks.yaml` alone for my non-nix-using colleagues?
I see that there is already [an issue](https://github.com/cachix/pre-commit-hooks.nix/issues/2) for documenting how to "override" a hook.  I also want to override a hook, but I think we mean "override" in different ways.

Most of my team doesn't use nix, so they run [pre-commit](https://pre-commit.com/) against a `.pre-commit-hooks.yaml` which I would be frowned at if I deleted and `.gitignored` as this warning recommends:

```
pre-commit-hooks.nix: WARNING: Refusing to install because of pre-existing .pre-commit-config.yaml
    1. Translate .pre-commit-config.yaml contents to the new syntax in your Nix file
        see https://github.com/cachix/pre-commit-hooks.nix#getting-started
    2. remove .pre-commit-config.yaml
    3. add .pre-commit-config.yaml to .gitignore
```

I maintain a flake.nix in my team's repo because I appreciate the determinism, especially when switching between machines.  This comes with certain headaches like [this one](https://github.com/astral-sh/ruff-pre-commit/issues/22).  Since `charliermarsh/ruff-pre-commit` breaks on NixOS but works for my non-nix coworkers, I'd like to do the following.

A: contribute a `ruff` pre-commit hook to pre-commit-hooks.nix
B: configure pre-commit-hooks.nix to run the nix-friendly hook instead of `charliermarsh/ruff-pre-commit`, but otherwise conform to the existing hooks 
C: commit this configuration to my repo in a way that does not break my non-nix coworker's workflows

So when I say "override" I mean:

> leave the traditional pre-commit configuration intact but have `pre-commit-hooks.nix` "take over" for certain hooks. 

Can I get some advice on whether pre-commit-hooks.nix is capable of supporting B and C?  If so, how would I configure it?

diefans:
This fixes #345, by adding a `configFile` option to change the default (`.pre-commit-config.yaml`) and so enables the parallel usage of devenv and non-devenv maintained pre-commit hooks.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -362,6 +362,12 @@ in
           display a warning message when a renamed option is used.
         '';
       };
+
+      configFile = lib.mkOption {
+        type = types.str;
+        default = ".pre-commit-config.yaml";
+        description = ''The name of the config file.'';
+      };
     };
 
   config =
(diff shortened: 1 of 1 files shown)


sandydoo:
1. There are a few more `pre-commit` calls in `run` that would need to be adjusted. https://github.com/cachix/git-hooks.nix/blob/6d7586d0d35ccd1e87bb0ffb7c0d38b4cc0fe469/modules/pre-commit.nix#L55-L83
2. I'm not sure about the name of the option. We may also need to expose `configFile` itself in the future, e.g. #494. Maybe `configPath` would be better?

diefans:
@sandydoo ...I am full of work at the moment, but this topic is not forgotten!

willfish:
This is a make or break option for me at work since I work with pre-existing pre-commit configuration and I can't just delete/.gitignore that. I'm using flakes in about 10 repos that I can just add this tooling for and make my life a bunch easier

willfish:
Absolutely love this project <3

PR #499, opened 2024-10-01
Add extraArgs for clippy

mkenigs:
I would like to pass `--tests` to clippy. I figured rather than adding an option just for tests, it might be better to just add an option to pass through any arguments


mkenigs:
Thanks!

sivizius:
Related to https://github.com/cachix/git-hooks.nix/pull/434

PR #501, opened 2024-10-06
add golines pre-commit hook

hmajid2301:
I tried to copy the gofmt and move it over to golines as the tooling is very similar. 
https://github.com/segmentio/golines

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -565,6 +565,20 @@ in
           };
         };
       };
+      golines = mkOption {
+        description = "golines hook";
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.settings = {
+            flags = mkOption {
+              type = types.str;
+              description = "Flags passed to golines. See all available [here](https://github.com/segmentio/golines?tab=readme-ov-file#options)";
+              default = "";
+              example = "-m 120";
+            };
+          };
+        };
+      };
       headache = mkOption {
         description = "headache hook";
         type = types.submodule {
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -2487,6 +2501,31 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
         # all file names in a single run.
         require_serial = true;
       };
+      golines =
+        {
+          name = "golines";
+          description = "A golang formatter that fixes long lines";
+          package = tools.golines;
```
sandydoo, line 2508: You have to import `golines` from pkgs and add it to tools. See `tools.nix`.

sandydoo, line 2508: You have to import `golines` from pkgs and add it to tools. See `tools.nix`.

sandydoo (approved):

sandydoo (changes requested):
You have to import `golines` from pkgs and add it to tools. See `tools.nix`.

hmajid2301:
I have made the change, please let me know if I did it correctly? Thanks

modules/hooks.nix
```diff
@@ -565,6 +565,20 @@ in
           };
         };
       };
+      golines = mkOption {
+        description = "golines hook";
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.settings = {
+            flags = mkOption {
```
sandydoo, line 573: For what it's worth, all hooks accept an `args` option that does this.

hmajid2301: ahh good to know for the next one thanks

PR #506, opened 2024-10-11
statix: fix broken list option `--ignore`

### LINKED ISSUE #505: Option names passed to `statix` are missing dashes
Changes made in #503 introduced a call to `lib.cli.toGNUCommandLineShell` but it does not call `mkOption` on `name`s so they are without dashes. I have reproduced the issue with expression
```nix
{
  pkgs ? import <nixpkgs> { },
}:
let
  inherit (pkgs) lib;
  settings = {
    format = [ "errfmt" ];
    ignore = [ "hardware-config.nix" ];
  };
in
{
  goodOptions = lib.cli.toGNUCommandLineShell { } settings;
  badOptions = lib.cli.toGNUCommandLineShell {
    mkList = name: value: [ name ] ++ lib.unique value;
  } settings;
}
```
.
Here `goodOptions` evaluates to `'--format' 'errfmt' '--ignore' 'hardware-config.nix'` and `badOptions` to `'format' 'errfmt' 'ignore' 'hardware-config.nix'`.

sandydoo:
Fixes #505.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -3392,9 +3392,15 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
           entry =
             let
               inherit (hooks.statix) package settings;
+              mkOptionName = k:
+                if builtins.stringLength k == 1
+                then "-${k}"
+                else "--${k}";
               options = lib.cli.toGNUCommandLineShell
                 {
-                  mkList = name: value: [ name ] ++ lib.unique value;
+                  # instead of repeating the option name for each element,
+                  # create a single option with a space-separated list of unique values.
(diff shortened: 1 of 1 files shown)


modules/hooks.nix
```diff
@@ -3392,9 +3392,15 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
           entry =
             let
               inherit (hooks.statix) package settings;
+              mkOptionName = k:
+                if builtins.stringLength k == 1
+                then "-${k}"
+                else "--${k}";
```
trueNAHO, line 3398: Why are we re-implementing the default behaviour: https://github.com/NixOS/nixpkgs/blob/nixos-24.05/lib/cli.nix#L67-L73

PR #509, opened 2024-10-11
Another attempt to fix broken statix (fixes #508)

### LINKED ISSUE #508: statix is broken
Regression: statix `--ignore` option is generated without any value:

```json file:

143           "always_run": false,
144           "args": [],
145           "entry": "/nix/store/7vl6g9hb48ffdqkyx0dqmxzh1smng7yd-statix-0.5.8/bin/statix check --format errfmt --ignore",
146           "exclude": "^$",
147           "exclude_types": [],
...
```

in this particular case given the config:

```nix
git-hooks.lib.${system}.run {
  inherit src;

  hooks = {
    statix.enable = true;
  };

  tools = pkgs;
}
```

the `--ignore` option simply should not have been added at all.

ink-splatters:
One of commits introduced changes in how `statix` cli options are generated.

Fix attempt #506 was merged however it fails to resolve the issue, also, IMHO, adding unnecessary complexity, duplicating [lib.cli](https://github.com/NixOS/nixpkgs/blob/13e7885af14a46620124d2eb4008dfae3e2cfdf2/lib/cli.nix#L119) functionality

@sandydoo as #506 author, feel free to kindly comment if those changes are fine with you.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -3479,7 +3479,7 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
               inherit (hooks.statix) package settings;
               options = lib.cli.toGNUCommandLineShell
                 {
-                  mkList = name: value: [ name ] ++ lib.unique value;
+                  mkList = name: value: if value == [] then [] else [ name ] ++ lib.unique value;
                 }
                 settings;
             in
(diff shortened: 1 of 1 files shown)


sandydoo:
Oh, the empty case was broken. But this is now equally broken for the non-empty case. We still need the reverted commit, in addition to an empty check.

Maybe we should just go back to `mkCmdArgs` until this function matures.
And, oh god, do we need a test harness 😂

sandydoo:
Merged #510 with the empty-case fix. 

We still need `mkOptionName` for the option and `lib.cli` doesn't expose it, so we're stuck with this.

PR #512, opened 2024-10-11
feat(run.nix): expose more

sivizius:
With this PR, the derivation returned by `git-hooks.lib.${system}.run { … }` contains the attribute `config`, which can be used to get the configured pre-commit-hooks. I.e. t is now possible to use the result both in the flake-outputs as e.g. `checks = { inherit pre-commit-hooks; };` as well as `apps = fooBar pre-commit-hooks.config.hooks`, where `fooBar` extracts all entries of enabled hooks so e.g. `nix run clippy` would be possible using the same CLI-options as the `checks.pre-commit-hooks`. 

If there already was a way to access these processed hooks, please document it, because I could not find a way myself.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,5 +55,7 @@
             exposed-stable.checks)));
 
       lib = forAllSystems ({ exposed, ... }: { inherit (exposed) run; });
+
+      exposed = forAllSystems ({ exposed, ... }: exposed);
     };
 }
--- a/nix/run.nix
+++ b/nix/run.nix
@@ -35,6 +35,7 @@ let
 
 in
 project.config.run // {
+  inherit (project) config;
+  inherit (project.config) enabledPackages;
   shellHook = installationScript;
-  enabledPackages = project.config.enabledPackages;
 }


trueNAHO:
> If there already was a way to access these processed hooks, please document it, because I could not find a way myself.

Does a `self` access resolve this:

```nix
checks.git-hooks = inputs.git-hooks.lib.${system}.run {};
apps.default = self.checks.${system}.git-hooks;
```

However, I agree that being able to access the internal `hooks` values would be a great feature addition.

For example, my LaTeX flake requires me to wrap `hooks.latexindent.package` in `hooks.latexindent.entry` because I need to be able to evaluate the `"$(mktemp --directory)"` subshell, which is impossible with just `entry = "<COMMAND> $(<SUB_COMMAND>)'"`:

```nix
checks.git-hooks = inputs.git-hooks.lib.${system}.run {
  hooks.latexindent = {
    enable = true;

    entry = lib.getExe (
      pkgs.writeShellApplication {
        name = "git-hooks-latexindent-entry";

        text = ''
          ${
            lib.getExe'
            (
              # Re-implement and re-evaluate hooks.latexindent.package.
              pkgs.texlive.combine {
                inherit (pkgs.texlive) latexindent scheme-minimal;
              }
            )
            "latexindent"
          } ${
            lib.cli.toGNUCommandLineShell {} {
              logfile = toString /dev/null;
              overwriteIfDifferent = true;
              silent = true;
              yaml = "defaultIndent: '    '";
            }
          } --cruft "$(mktemp --directory)" "$@"
        '';
      }
    );
  };

  src = ./.;
};
```

Accessing the package through something like `hooks.latexindent.package` would be more convenient and would even benefit from the performance optimization from https://github.com/cachix/git-hooks.nix/pull/513.

PR #518, opened 2024-10-17
feat: add goimports as a hook

hmajid2301:
Added go imports, https://pkg.go.dev/golang.org/x/tools/cmd/goimports as a hook

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -563,6 +563,20 @@ in
           };
         };
       };
+      goimports = mkOption {
+        description = "goimports hook";
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.settings = {
+            flags = mkOption {
+              type = types.str;
+              description = "Flags passed to goimports.";
+              default = "";
+              example = "-local github.com/example/repo -w .";
+            };
+          };
+        };
+      };
       golines = mkOption {
         description = "golines hook";
         type = types.submodule {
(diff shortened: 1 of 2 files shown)


modules/hooks.nix
```diff
@@ -2598,6 +2612,31 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.ormol
...
+          description = "Go import lines, adding missing ones and removing unreferenced ones.";
+          package = tools.gotools;
+          entry =
+            let
+              script = pkgs.writeShellScript "precommit-goimports" ''
+                set -e
+                failed=false
+                for file in "$@"; do
+                    # redirect stderr so that violations and summaries are properly interleaved.
+                    if ! ${hooks.gotools.package}/bin/goimports ${hooks.goimports.settings.flags} -w "$file" 2>&1
```
hmajid2301, line 2627: Is there a way to do with args instead of manually specifying flags here and setting them above?

sandydoo: Don't think so, since `$@` is already used for the files.

sandydoo: If the tool is not too complicated, the general pattern is that we add dedicated options to make them show up in the docs of downstream projects (flake-parts, devenv). 
Some hooks will have an `extraFlags`/`extraOptions` setting to capture everything. Use the existing hooks for inspiration.

nix/tools.nix
```diff
@@ -85,6 +85,7 @@
 , go-tools
 , golangci-lint
 , golines
+, gotools
```
sandydoo, line 88: We need to make sure we can evaluate this on older nixpkgs, as this tool is quite new to nixpkgs.

```suggestion
, gotools ? null
```

sandydoo (changes requested):

PR #522, opened 2024-10-23
fix: avoid use of deprecated stage names

r-vdp:
These cause warnings with recent versions of pre-commit, and will be
removed in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ nix develop
       pre-commit-check = nix-pre-commit-hooks.run {
         src = ./.;
         # If your hooks are intrusive, avoid running on each commit with a default_states like this:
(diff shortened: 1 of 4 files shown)


trueNAHO (approved):
I have successfully tested that this PR eliminates the following pre-commit hook warning when running for example `pre-commit run --all-files`:

```
[WARNING] hook id `<HOOK>` uses deprecated stage names (commit) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
```

Note that running the upstream suggested `pre-commit migrate-config` command does not resolve the git-hooks.nix issue.

sandydoo:
Thanks!

PR #523, opened 2024-11-04
fix: poetry option is not defined

sheeeng:
`error: The option 'pre-commit.hooks.poetry.entry' is used but not defined.`


loicreynier:
You should use `hooks.poetry-check` and/or `hooks.poetry-lock` when you want to run `poetry check` or `poetry lock`, see [`hooks.nix`](https://github.com/loicreynier/git-hooks.nix/blob/622291c026190caf13cb26f5136616b1ff0a07aa/modules/hooks.nix#L2964-L2979).

PR #528, opened 2024-11-20
Run shellcheck on installation hook tests

nicknovitski:
Because of the way the tests are written, this also shellchecks each test derivation's shellHook, which gets a lot of coverage of the shellHooks in general.

Besides the obvious bash-level fixes, some of the warnings and errors I fixed by moving the logic out into nix, which felt a lot simpler.

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -2,12 +2,13 @@
 let
   inherit (lib)
     boolToString
+    concatMapStrings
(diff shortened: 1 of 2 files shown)


nicknovitski:
Rebased on master.  I've also run the flake checks on `aarch64-darwin`.

nicknovitski:
Rebased and tested again.

PR #531, opened 2024-12-09
feat: add an option to `cabal2nix` hook for specifying output filename.

rsrohitsingh682:
At present after enabling `cabal2nix` hook, it generates a file named `default.nix` which is hardcoded. It is better if user has an option to specify the filename instead of using hardcoded `default.nix`.

This PR adds an option to `cabal2nix` hook to specify output filename generated by running `cabal2nix`.

Example: 

```
cabal2nix = {
   enable = true;
   settings.output_filename = "cabal.nix";
};
```

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -213,7 +213,7 @@ in
         type = types.submodule {
           imports = [ hookModule ];
           options.settings = {
-            output_filename =
+            outputFilename =
               mkOption {
                 type = types.str;
                 description = "The name of the output file generated after running `cabal2nix`.";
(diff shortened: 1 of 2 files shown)


rsrohitsingh682:
@sandydoo @NorFairKing Any thoughts on this PR ?

NorfairKing:
I don't see a use case so think this needlessly complicating things but I also don't really care :shrug:

rsrohitsingh682:
> I don't see a use case so think this needlessly complicating things but I also don't really care 🤷

It's required for me as I don't want pre-commit hook to generate a `default.nix`.  Anyways since the default is `default.nix` so the default behaviour still remains the same.

rsrohitsingh682:
@sandydoo Could you please help with the review ?

modules/hooks.nix
```diff
@@ -208,6 +208,20 @@ in
           };
         };
       };
+      cabal2nix = mkOption {
+        description = "cabal2nix hook";
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.settings = {
+            output_filename =
```
sandydoo, line 216: Lets stick with camelCase for the settings, if possible.

sandydoo (changes requested):
Otherwise, LGTM

sandydoo:
The CI failure is unrelated. This branch needs a rebase on master.

sandydoo (approved):

PR #533, opened 2024-12-13
feat: ordering in hooks.

### LINKED ISSUE #250: Hook order ignored
Follow instructions from readme: https://github.com/cachix/pre-commit-hooks.nix/blob/61a3511668891c68ebd19d40122150b98dc2fe3b/README.md?plain=1#L31-L73

Now inspect `.pre-commit-config.yaml`:

<details>

```yaml
# DO NOT MODIFY
# This file was generated by pre-commit-hooks.nix
{
  "default_stages": [
    "commit"
  ],
  "repos": [
    {
      "hooks": [
        {
          "entry": "/nix/store/yyc8pmfy43nl2hwfb3n16rcd866nqki8-elm-format-0.8.5/bin/elm-format --yes --elm-version=0.19",
          "exclude": "^$",
          "files": "\\.elm$",
          "id": "elm-format",
          "language": "system",
          "name": "elm-format",
          "pass_filenames": true,
          "stages": [
            "commit"
          ],
          "types": [
            "file"
          ],
          "types_or": []
        },
        {
          "entry": "/nix/store/gvp4jil2i6r4n7kvx4sh4d7vhp5v433j-ormolu-0.3.1.0-bin/bin/ormolu --mode inplace '--ghc-opt' '-Xlhs' '--ghc-opt' '-Xhs' ",
          "exclude": "^$",
          "files": "\\.l?hs(-boot)?$",
          "id": "ormolu",
          "language": "system",
          "name": "ormolu",
          "pass_filenames": true,
          "stages": [
            "commit"
          ],
          "types": [
            "file"
          ],
          "types_or": []
        },
        {
          "entry": "/nix/store/25i9mjmk2a0pq19jgiscn3wfc539ya3a-shellcheck-0.8.0-bin/bin/shellcheck",
          "exclude": "^$",
          "files": "",
          "id": "shellcheck",
          "language": "system",
          "name": "shellcheck",
          "pass_filenames": true,
          "stages": [
            "commit"
          ],
          "types": [
            "shell"
          ],
          "types_or": [
            "sh",
            "ash",
            "bash",
            "bats",
            "dash",
            "ksh"
          ]
        }
      ],
      "repo": "local"
    }
  ]
}
```

</details>

You'll see the order of hooks is:

- elm-format
- ormolu
- shellcheck

Now, change the order of hooks in `default.nix`. Set them like:

```nix
    hooks = {
      shellcheck.enable = true;
      ormolu.enable = true;
      elm-format.enable = true;
    };
```

Again, enter `nix-shell` and inspect `.pre-commit-config.yaml`. You'll see that the hooks are in the same order as the 1st time. The file is in fact the same.

This is not a big problem in this case, but in other cases, order matter.

For example, if you're writing python and want to enable `autoflake` and `flake8`, you will most likely want `autoflake` to run before `flake8`, so `flake8` doesn't catch the errors that `autoflake` already fixed.

In specific cases like this one, we could still provide a sane default order. But there are custom hooks, and we can't predict in which order they should be executed.

So, this project needs IMHO a way to specify hook order, just like normal pre-commit does.

rsrohitsingh682:
resolves #250

### DIFF
--- a/modules/hook.nix
+++ b/modules/hook.nix
@@ -135,15 +135,6 @@ in
       default = true;
     };
 
-    priority = mkOption {
-      type = types.number;
-      description = ''
-        Defines the priority order in which hooks are executed.
-        Lower the number, the higher the precedence.
-      '';
-      default = 0;
-    };
-
     fail_fast = mkOption {
       type = types.bool;
       description = ''
@@ -193,12 +184,31 @@ in
         '';
       default = [ ];
     };
+
+    before = mkOption {
+      type = types.listOf types.str;
+      description =
+        ''
+          List of hooks that should run after this hook.
+        '';
+      default = [ ];
+    };
+
+    after = mkOption {
+      type = types.listOf types.str;
+      description =
+        ''
+          List of hooks that should run before this hook.
+        '';
+      default = [ ];
+    };
+
   };
 
   config = {
     raw =
       {
-        inherit (config) name entry language files types types_or exclude_types pass_filenames priority fail_fast require_serial stages verbose always_run args;
+        inherit (config) name entry language files types types_or exclude_types pass_filenames fail_fast require_serial stages verbose always_run args before after;
         id = config.name;
         exclude = mergeExcludes config.excludes;
       };
--- a/modules/hook.nix
+++ b/modules/hook.nix
@@ -135,6 +135,15 @@ in
       default = true;
     };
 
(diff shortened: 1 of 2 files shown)


domenkozar:
I'd very much prefer `before/after` solution of topological ordering. There's already examples of this in NixOS for inspiration!

PR #535, opened 2024-12-15
bugfix: remove before/after options from config json file

### LINKED ISSUE #112: [WARNING] Unexpected key(s) present on local => nixpkgs-fmt: types_or
With nixos-unstable, I'm getting warnings like these

```
[WARNING] Unexpected key(s) present on local => nixpkgs-fmt: types_or
[WARNING] Unexpected key(s) present on local => ormolu: types_or
[WARNING] Unexpected key(s) present on local => shellcheck: types_or
```

It appears that the configuration format has changed with a pre-commit update, or at least it has started checking for unknown fields.

djacu:
A recent commit to add ordering priority to the hooks introduced new keys into the pre-commit JSON config file. These keys should not be there are they are not keys pre-commit is aware of and produce warning messages. e.g.:

```
[WARNING] Unexpected key(s) present on local => treefmt: after, before
```

This commit fixes it by filter out all keys with names `before` or `after` before writing the configuration to file.

Fixes #112


djacu:
I'll note that this method for fixing is a bit ham-fisted and will delete any key with name `before` or `after`. It does produce the intended result of fixing the related issue, however, a more precise method would be better.

djacu:
Closing in favor of https://github.com/cachix/git-hooks.nix/pull/536

PR #536, opened 2024-12-15
bugfix: remove before/after options from config json file

djacu:
A recent commit to add ordering priority to the hooks introduced new keys into the pre-commit JSON config file. These keys should not be there are they are not keys pre-commit is aware of and produce warning messages. e.g.:

```
[WARNING] Unexpected key(s) present on local => treefmt: after, before
```

This commit fixes it by filter out all keys with names `before` or `after`, in Nix, before writing the configuration to file.

Fixes https://github.com/cachix/git-hooks.nix/issues/112

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -9,6 +9,7 @@ let
     mapAttrsToList
     mkOption
     types
+    removeAttrs
     remove
     ;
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Thank you!

PR #548, opened 2025-01-16
Fix Terraform validate missing init on subdirectory

shikanime:


sandydoo:
@shikanime, is this necessary for the `fmt` command?

shikanime:
If I have a repository that has terraform module in subdirectory and the terraform fmt is runt in the root, it will complain that the current directory is not a terraform module such as:

```
root/ # Not a Terraform module
├─ envs/
   ├─ prod/ # Terraform module
   │  ├─ *.tf
   ├─ dev/ # Terraform module
      ├─ *.tf
```

sandydoo:
What if we use the `-recursive` flag?

sandydoo:
> If I have a repository that has terraform module in subdirectory and the terraform fmt is runt in the root, it will complain that the current directory is not a terraform module such as:

Thanks for the repro. I'll look into this.

sandydoo:
I can't 

> If I have a repository that has terraform module in subdirectory and the terraform fmt is runt in the root, it will complain that the current directory is not a terraform module such as:
> 
> ```
> root/ # Not a Terraform module
> ├─ envs/
>    ├─ prod/ # Terraform module
>    │  ├─ *.tf
>    ├─ dev/ # Terraform module
>       ├─ *.tf
> ```

I can't seem to replicate this. Both terraform and opentofu are happy to format the passed in files from any directory.

shikanime:
On second thought, oh my mistake, it was just the validate command that was complaining, sorry 🙏

PR #563, opened 2025-03-01
Add a `configPath` option

### LINKED ISSUE #345: [question] Can I use this to override a hook while leaving `pre-commit-hooks.yaml` alone for my non-nix-using colleagues?

sandydoo:
Adds a `configPath` option to allow changing the path to the generated pre-commit config (`.pre-commit-config.yaml` by default). 

Fixes #345.

Based on #496.


willfish:
Yay. Thanks for doing this

zevisert-aaico:
Hey @sandydoo thanks for this! I've been annoyed by #345 for a while as well.. As a cachix/devenv user, would this work out of the box with devenv, or does devenv need a PR to enable this? I tried the following in my devenv.nix, but it doesn't seem to be working (no pre-commit configuration is created at `.devenv/state/.pre-commit-config.yaml`):

```nix
{
  config,
  ...
}: {
 git-hooks = {
    configPath = builtins.toPath (config.devenv.state + "/.pre-commit-config.yaml");
    hooks = {
      trim-trailing-whitespace.enable = true;
    };
  };
}
```

zevisert:
Hey @sandydoo thanks for this! I've been annoyed by #345 for a while as well.. As a cachix/devenv user, would this work out of the box with devenv, or does devenv need a PR to enable this? I tried the following in my devenv.nix, but it doesn't seem to be working (no pre-commit configuration is created at `.devenv/state/.pre-commit-config.yaml`):

```nix
{
  config,
  ...
}: {
  git-hooks = {
    configPath = builtins.toPath (config.devenv.state + "/.pre-commit-config.yaml");
    hooks = {
      trim-trailing-whitespace.enable = true;
    };
  };
}
```

sandydoo:
@zevisert the path gets prefixed by the git working directory. We should make a note about this until we can support absolute paths as well. You can use a string for now, like `".devenv/state/.pre-commit-config.yaml"`.

zevisert:
Ah, that was it - thank you!

PR #566, opened 2025-03-06
docs: add supported but non-mentioned hooks

f1rstlady:
- **docs: sort hook sections by name**
- **docs: add supported but non-mentioned hooks**


sandydoo:
Great stuff!

PR #578, opened 2025-03-15
Add detect-secrets as a new git hook

onnimonni:
Hey,

I would want to add support for https://github.com/Yelp/detect-secrets into devenv.


onnimonni:
In the hindsight the already currently supported secret checker tools seem to be sufficient for me and thus I will close this PR.

I have set them up here if someone wants to see example:
https://github.com/midworkhq/midwork/blob/f83e2a2db635ee9569eab5b3258cbed42e6d6607/devenv.nix#L62-L71

PR #583, opened 2025-04-13
typos improvements

bmrips:
This PR comprises three changes:

- Use structured configuration by making `typos.settings.configuration` an attrs. This change is not backwards compatible.
- Add the `typos.settings.force-exclude` option that maps to typos' `--force-exclude` argument. In addition, enable it by default since you expect exclude globs set via `--exclude` CLI option or the configuration file's `[files.extend-exclude]` option to be effective.
- Allow multiple exclusion patterns in `typos.settings.exclude`.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1831,10 +1831,10 @@ in
 
             exclude =
               mkOption {
-                type = types.str;
-                description = "Ignore files and directories matching the glob.";
-                default = "";
-                example = "*.nix";
+                type = with types; coercedTo str (s: [ s ]) (listOf str);
+                description = "Ignore files and directories matching one of the globs.";
+                default = [ ];
+                example = [ "*.nix" ];
               };
 
             force-exclude =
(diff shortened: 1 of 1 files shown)


srid:
I find a strange behaviour when using this PR.

For this setting,

```nix
          settings.exclude = [
            "*.nix"
            ".hlint.yaml"
          ];
```

it generates a YAML with the following CLI:

```
❯ rg bin/typos .pre-commit-config.yaml
101:          "entry": "/nix/store/n97528zsiq2yshvfq5whl4xpbydlnddl-typos-1.32.0/bin/typos --config /nix/store/0my2fa9n84gy2jk0wlclww22xzwckw0l-typos-config.toml --exclude *.nix  --force-exclude",
```

Note that `--exclude *.nix` part. Somehow only the first exclude entry (`*.nix`) is being used, the rest (`.hlint.yaml`) is ignored.

But I'd think that the following use of `map` *should* have included all entries:

```nix
excludeFlags = lib.map (glob: "--exclude ${glob}") exclude;
```

Is this something to do with the use of `coercedTo`?

bmrips:
> Note that `--exclude *.nix` part. Somehow only the first exclude entry (`*.nix`) is being used, the rest (`.hlint.yaml`) is ignored.

Thank you for the notification! It turned out that `mkCmdArgs` only considers the first and second items of the list, such that following items are ignored. Should be fixed now.

lafrenierejm (approved):

PR #591, opened 2025-05-21
Fix (trufflehog): Remove renundant `--no-update` causing it to fail

Michael-C-Buckley:
Trufflehog was failing due to multiple no update flags being passed.

Fix: remove the `--no-update` from the hook to remove the redundantly sourced no update prompt.

```
trufflehog...............................................................Failed
- hook id: trufflehog
- exit code: 1

trufflehog: error: flag 'no-update' cannot be repeated, try --help
```


totoroot:
Should be fixed in #594.

PR #603, opened 2025-06-30
fix(pretty-format-json): make autofix default true

### LINKED ISSUE #602: pre-commit-hooks for pretty-format-json default behavior is misleading
**Describe the bug**
If I enable the hooks:

```nix
 git-hooks.hooks = {
    pretty-format-json.enable = true;
    check-json.enable = true;
  };
```

It will detect unformatted json but the file is not formatted after run.

**To reproduce**

Do what is in the description with any unformatted json.

If you change `settings.autofix = true` then it will behave as the original pre-commit hook.

In my opinion this should should not exist and be always `true`, it is `false` by default but a formatter is useless if it is not autofixing.

**Version**
devenv 1.6.1

staticdev:
Auto-explanatory title, the default behavior on pre-commit is autofix,

Closes #602

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -1327,7 +1327,7 @@ in
                 mkOption {
                   type = types.bool;
                   description = "Automatically format JSON files.";
-                  default = false;
+                  default = true;
                 };
               indent =
                 mkOption {


staticdev:
@sandydoo after further testing I verified you are absolutely RIGHT. Sorry for that, closing the issue and PR.

Issue #618, opened 2025-08-20
Typos exclude param implementation failing for globs


norbertwnuk:
There are two issues with current implementation of the exclude parameter for [the typos hook](https://github.com/cachix/git-hooks.nix/blob/master/modules/hooks.nix#L4027):
- the value should be passed in the apostrophes `"--exclude '${exclude}' --force-exclude"`, otherwise it is expanded by the shell and not working when more than one file is found
- the additional excludes property (to be backward compatible) makes more sense since more than one glob can be defined with multiple `--exclude` CLI params: ` /nix/store/xxx-typos-1.32.0/bin/typos --config /nix/store/yyy-typos-config.toml --exclude '*.json1' --exclude '*.json' --exclude '*.json2' --force-exclude`, the result configuration is more clear than for `--exclude '*.json1 *.json *.json2'` and both works:

```
 /nix/store/xxx-typos-1.32.0/bin/typos --config /nix/store/yyy-typos-config.toml --exclude '*.json1 *.json *.json2' --force-exclude --dump-config - | grep exclude -A5
extend-exclude = [
    "**.png",
    "*.json1 *.json *.json2",
]

/nix/store/xxx-typos-1.32.0/bin/typos --config /nix/store/yyy-typos-config.toml --exclude '*.json1' --exclude '*.json' --exclude '*.json2' --force-exclude --dump-config - | grep exclude -A5
extend-exclude = [
    "**.png",
    "*.json1",
    "*.json",
    "*.json2",
]
```

All those observations empirically tested, implementation can be checked here: https://github.com/crate-ci/typos/blob/master/crates/typos-cli/src/bin/typos-cli/args.rs#L216

PR #619, opened 2025-08-20
Fix exclude param propagation for typos hook

norbertwnuk:
Simple fix for: Typos exclude param implementation failing for globs #618


sandydoo:
Thanks! Cherry-picked in 59cc72533edf0788599ea312198eeeca2c48e931 due to conflicts.

PR #627, opened 2025-09-15
feat: nixfmt indent

shimeoki:
nixfmt 1.0.0 added an `--indent=INT` option to configure indentation width. this PR adds this option to `nixfmt-rfc-style` settings to avoid using `args = [ "--indent=X" ]`.

aside from that, i didn't like how options were handled previously, so i refactored it a bit. if this is too much (because the solution is a bit hacky and it's still not DRY), only the first commit with the feature itself without any refactoring is available - i can do a force push to it.

also, the indent option description should probably be `"Indentation width."`. noticed that only while writing the PR description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@
 
 - Run hooks **as part of development** and **during CI**
 
+- Support for alternative `pre-commit` implementations, like [prek](https://github.com/j178/prek).
+
 ## Getting started
 
 ### devenv.sh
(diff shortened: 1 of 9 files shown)


shimeoki:
sorry for the late reply.

- added the indent option for the `nixfmt` as well
- changed the descriptions for the options in `nixfmt` and `nixfmt-rfc-style` to match the cli help
- used `mkCmdArgs`

for both `nixfmt` and `nixfmt-rfc-style` i did a check for below `1.0.0` to throw an error if indent is used.

also, i am not very sure about this, but i removed the `visible = false` for the `nixfmt` hook. i think the visibility was off because it was deprecated, but i suppose it's not the case anymore.

PR #634, opened 2025-09-29
Support alternative `pre-commit` implementations like `prek`

leiserfg:
I replaced the explicit calls to `pre-commit` with `getExe` so there is a single `if` involved. I didn't update the readme and docs references to pre-commit 'cause  I'm not sure how we wanna handle it.

### DIFF
--- a/nix/run.nix
+++ b/nix/run.nix
@@ -18,7 +18,7 @@ let
             {
               _module.args.pkgs = pkgs;
               _module.args.gitignore-nix-src = gitignore-nix-src;
-              package = lib.mkDefault (if builtins.hasAttr "prek" pkgs then pkgs.prek else pkgs.pre-commit);
+              package = lib.mkDefault pre-commit;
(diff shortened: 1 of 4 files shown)


nix/run.nix
```diff
@@ -18,7 +18,7 @@ let
             {
               _module.args.pkgs = pkgs;
               _module.args.gitignore-nix-src = gitignore-nix-src;
-              package = lib.mkDefault pre-commit;
+              package = lib.mkDefault (if builtins.hasAttr "prek" pkgs then pkgs.prek else pkgs.pre-commit);
```
sandydoo, line 21:
```suggestion
              package = lib.mkDefault pre-commit;
```

sandydoo (changes requested):
This looks great!

I'd like to merge this without changing the default to `prek` for now.
After a period of testing (perhaps in devenv), we can make the switch in a separate PR.

PR #635, opened 2025-09-30
Added hooks.golangci-lint.settings.flags option

tobias-kuendig:
It is currently not possible to define any of the many golangci-lint flags: https://golangci-lint.run/docs/configuration/cli/#run 

This PR adds the option to do so.


sandydoo:
Does `hooks.golangci-lint.args = [ ... ];` work for this hook?

PR #643, opened 2025-10-18
gitlint: add `flags` option

### LINKED ISSUE #641: gitlint: error on using "args"
When using the gitlint commit hook with custom arguments, then you will get an error message like:

```
Error: Invalid value for '--msg-filename': '-c': No such file or directory
```

#### Example

```nix
git-hooks.hooks.gitlint.args = ["-c" "general.ignore=B6" "-c" "title-max-length.line-length=120"];
```

#### Workaround

Overriding `entry` like:

```nix
git-hooks.hooks.gitlint.entry = "${pkgs.gitlint}/bin/gitlint -c general.ignore=B6 -c title-max-length.line-length=120 --staged --msg-filename";
```

@bmrips

bmrips:
The gitlint hook requires `--msg-filename` to appear last on the command line such that it is in front of the commit message filename. This is not the case when the `args` option is set. To fix this, we append `--msg-filename` to the `args` option.

Fixes #641.

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -639,6 +639,18 @@ in
           };
         };
       };
+      gitlint = mkOption {
+        description = "gitlint hook";
+        type = types.submodule {
+          imports = [ hookModule ];
+          options.flags = mkOption {
+            type = types.listOf types.str;
+            description = "Flags passed to gitlint. See all available [here](https://jorisroovers.com/gitlint/latest/configuration/cli/)";
+            default = [ ];
+            example = [ "-c" "title-max-length.line-length=120" ];
+          };
+        };
+      };
       golines = mkOption {
         description = "golines hook";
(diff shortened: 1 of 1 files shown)


sandydoo (changes requested):
`args` are for end users to set. If a hook is designed in such a way that it cannot handle filenames as arguments, ie. without an option like `--msg-filename`, then we usually expose a `settings.flags` or, better yet, individual options.

sandydoo:
> With a flags option, however, we loose forward compatibility.

Not entirely sure what you mean by that

bmrips:
> Not entirely sure what you mean by that

If we add `gitlint.flags.*` options and gitlint changes its CLI flags, we would have to keep them in sync.

bmrips:
Ok, I added the `gitlint.flags` option.

PR #651, opened 2025-11-10
Add nixf-diagnose

Stunkymonkey:
inlcude: https://github.com/inclyc/nixf-diagnose

i tested it manually in my flake.

it would be very helpfull, because nixpkgs currently runs this on every PR as well.


Stunkymonkey:
anything that is missing in this PR? @sandydoo

PR #668, opened 2025-12-10
feat(terraform-validate): support both opentofu and terraform

veselyn:
Inspired by https://github.com/cachix/git-hooks.nix/pull/419

### DIFF
--- a/modules/hooks.nix
+++ b/modules/hooks.nix
@@ -4130,17 +4130,17 @@ lib.escapeShellArgs (lib.concatMap (ext: [ "--ghc-opt" "-X${ext}" ]) hooks.fourm
           entry =
             let
               terraform-validate = pkgs.writeScriptBin "terraform-validate" ''
-              #!/usr/bin/env bash
-              set -x
-              for arg in "$@"; do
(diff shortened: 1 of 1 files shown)


nix/tools.nix
```diff
@@ -83,6 +83,7 @@
 , stylua
 , tagref
 , taplo
+, terraform
```
sandydoo: This needs to then be exported from the tools. I would double-check that it doesn't break eval for pkgs without unfree when _not_ using the hook.

sandydoo (changes requested):

nix/tools.nix
veselyn: Here's an example that works with OpenTofu and `allowUnfree: false`: https://github.com/veselyn/git-hooks.nix/tree/terraform-validate-package-test/test. Also no issues with evaluation with the hook disabled.

veselyn: Looks like it doesn't evaluate during the flake check after all.😄 I’ve removed it from the list, just like how it’s done in `terraform-format`.

PR #671, opened 2026-01-02
feat(modules/hook): change "system" language to "unsupported"

acid-bong:
`language: system` is getting deprecated and replaced with `language: unsupported` since Pre-commit v4.4.0[^1][^2]

[^1]: https://github.com/pre-commit/pre-commit/releases/tag/v4.4.0
[^2]: https://pre-commit.com/#unsupported

There's no warnings yet, but let's get ready just in case

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,9 +571,9 @@ Example configuration:
          excludes = [ "irrelevant\\.c" ];
 
          # The language of the hook - tells pre-commit
-         # how to install the hook (default: "system")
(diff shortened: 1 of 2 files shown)


acid-bong:
nah, just checked, Prek supports both old and new names: https://github.com/j178/prek/blob/df913baafec8189b0ccf736e4f8bf67b45959a36/crates/prek/src/config.rs#L95-L98

PR #677, opened 2026-01-18
Adds support for prek's priority and prek's builtin hooks

leiserfg:
The priorities requires a newer version of prek than it's available but soon. Built-ins are there already for a long time and more are added so make sense to leverage them.

### DIFF
--- a/modules/hook.nix
+++ b/modules/hook.nix
@@ -231,6 +231,17 @@ in
         '';
       default = [ ];
     };
+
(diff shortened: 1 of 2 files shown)


domenkozar:
Looks good to me, will let @sandydoo review this one.

philocalyst:
Drive by review: Looks great :)

leiserfg:
Done

PR #687, opened 2026-02-05
Reformat with nixfmt

mmlb:
nixpkgs-fmt is deprecated and recommends nixfmt, lets move over to it. I also created .git-blame-ignore-revs file and added the 2 previous format related commits it. I verified that there was no non-format related changes in them. I did so by going to previous commit, running pre-commit to get them re-formatted, then removing all whitespace and getting b3sum of content then comparing with same whitespace-less b3sum of the reformat commits. This way the ignored commits do not have any actual nix changes.


mmlb:
@zimbatm ping

PR #713, opened 2026-05-15
Extract prek builtin hooks

leiserfg:


modules/pre-commit.nix
```diff
@@ -36,7 +36,7 @@ let
         (builtins.attrValues enabledHooks);
     in
     if sortedHooks ? result then
-      builtins.map (value: value.raw) sortedHooks.result
```
domenkozar, line 39: won't this change make pre-commit bark on priority?

leiserfg: Let me change it to use `localHooks` there to be sure, thanks.

PR #714, opened 2026-05-17
feat: fall back to main worktree config in linked worktrees

### LINKED ISSUE #710: allow pre-commit checks conditional on their presence
i've found this repo quite useful for handling git hooks.
when using git worktrees however, i've found i wouldn't be able to normally interact with git (without `--no-verify`) until after opening up a nix shell to generate the hook files.
i tend to not really care to deal with the hooks in such cases, which has made me wonder if it might be possible to configure things such as to run them only when actually present already.
not sure if this is technically possible, or even deemed desirable - hence i figured i'd check up with y'all here.

KiaraGrouwstra:
Allows git worktrees to fall back to git hooks from their linked worktree.

Closes #710.

Assisted-by: Claude:claude-sonnet-4-6

### DIFF
--- a/modules/pre-commit.nix
+++ b/modules/pre-commit.nix
@@ -543,6 +543,37 @@ in
               # Convert the absolute path to a path relative to the toplevel working directory.
               common_dir=''${common_dir#''$GIT_WC/}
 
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra:
this has turned out not robust for my use-case

PR #721, opened 2026-06-17
inputs: drop gitignore.nix

rorosen:
Replaces `gitignore.nix` with `pkgs.nix-gitignore`, which should behave in the same way and reduce dependencies. Makes it also easier to import `nix/default.nix`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,26 +16,6 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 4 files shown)


sandydoo (approved):

sandydoo:
Nice!

